PR #36, opened 2015-01-05
Improve quality of coverage reports by parsing source files for functions

timholy:
This alters---and perhaps improves---our ability to quantify coverage. It leverages @jakebolewski's JuliaParser.jl to extract the beginning/end line range of "simple" function expressions. It then amends the coverage information by inserting 0 instead of `nothing` for functions that were never compiled.

This is not perfect. Here are deficits that I'm aware of currently:
- It misses functions defined inside a `for f in (:min, :max, ...) @eval begin...` style block (this error will contribute to overly-optimistic estimates of coverage)
- It falsely penalizes `end` statements and the initial `function f(...)` line in a multi-line function (this error will contribute to overly-pessimistic estimates of coverage)

There is also the genuine risk of the line count getting "out of sync", since the logic for extracting line numbers is not entirely trivial. I haven't seen it yet, but frankly I've only tested against a single source file.

I suspect both of these deficits may be fixable, so I am not certain this is yet worth merging. OTOH, I am not entirely certain I want to dump a huge amount of effort into this; if this is viewed as a step forward, perhaps we should go for it. But thanks to Jake it was surprisingly easy to get this far; maybe it wouldn't be so much harder to fix one or both of these problems (but I don't have any good ideas yet).

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,52 @@ Coverage.jl
 
 **"Take Julia test coverage results and do useful things with them."**
 
-Right now, that is submitting them to [Coveralls.io](https://coveralls.io), a test-coverage tracking tool that integrates with your continuous integration solution (e.g. [TravisCI](https://travis-ci.org/)).
+## Working locally
+
+### Code coverage
+
+Navigate to your test directory, and start julia like this:
+```sh
+julia --code-coverage=user
+```
+or, if you're running julia 0.4 or higher,
+```sh
+julia --code-coverage=user --inline=no
+```
+(Turning off inlining gives substantially more accurate results.)
+
+Then, run your tests (e.g., `include("runtests.jl")`) and quit julia.
+
+Finally, navigate to the top-level directory of your package, restart julia (with no special flags this time), and analyze coverage using
+```julia
+using Coverage
+covered, tot = coverage_folder()  # defaults to src/; alternatively, supply the folder name as a string
+```
+The fraction of total coverage is equal to `covered/tot`.
+
+### Memory allocation
+
+Start julia with
+```sh
+julia --track-allocation=user
+```
+Then:
+- Run whatever commands you wish to test. This first run is to ensure that everything is compiled (because compilation allocates memory).
+- Call `clear_malloc_data()` (or, if running julia 0.4 or higher, `Profile.clear_malloc_data()`)
+- Run your commands again
+- Quit julia
+
+Finally, navigate to the directory holding your source code. Start julia (without command-line flags), and analyze the results using
+```julia
+using Coverage
+analyze_malloc(dirnames)  # could be "." for the current directory, or "src", etc.
+```
+This will return a vector of `MallocInfo` objects, specifying the number of bytes allocated, the file name, and the line number.
+These are sorted in increasing order of allocation size.
+
+## Using Coveralls
+
+[Coveralls.io](https://coveralls.io) is a test-coverage tracking tool that integrates with your continuous integration solution (e.g. [TravisCI](https://travis-ci.org/)).
 
 ## Using Coverage.jl with Coveralls.io?
 
--- a/REQUIRE
+++ b/REQUIRE
@@ -2,3 +2,4 @@ julia 0.3-
 JSON
 Requests
 JuliaParser
+Compat
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.3-
 JSON
 Requests
+JuliaParser
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -5,10 +5,12 @@
 #######################################################################
 module Coverage
 
+    import JuliaParser.Parser
+
     # process_cov
     # Given a .cov file, return the counts for each line, where the
     # lines that can't be counted are denoted with a -1
-    export process_cov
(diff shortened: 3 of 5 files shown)


jakebolewski:
You would also have to turn off inlining, but this might be a step forward.

timholy:
We can just do that with `--inline=no` now.

IainNZ:
@timholy, with the second bullet point, does that apply to only functions that were not caught before, or does it apply to all functions?

IainNZ:
I'd like to try this out on JuMP, it seems like a great step forward.

timholy:
Looking at it a bit more, unfortunately the line numbering is borked. I think this needs to be fixed in JuliaParser.

timholy:
See https://github.com/jakebolewski/JuliaParser.jl/issues/13

timholy:
OK, I think I've fixed _both_ of the problems above. I hacked around the line number problem by first finding the file position corresponding to the beginning of each new source line, and then using that to keep track of where I am in the file. And it seems that with proper line number accounting, my concerns about `end`, etc, went away.

timholy:
One remaining problem is: what to do about whole _files_ that lack tests? In that case, there is no `*.cov` file.

timholy:
AFAICT, this will never pass travis; it seems that Pkg uses the old `REQUIRE` file to install dependencies, so `JuliaParser` does not get installed. Consequently, this PR fails its tests. You can, of course, try it locally (it works for me).

aviks:
+1000

This is a vastly better state of affairs than previously. I'd stopped looking at coverage results, since adding a single test gave you a 90% coverage statistics rate.

hayd:
@timholy couldn't you (as a bandaid) Pkg.add JuliaParser in the [.travis.yml](https://github.com/IainNZ/Coverage.jl/blob/be6484dcad8febce6f9bd491d49f8001171df291/.travis.yml#L18).

This is going to be ace!!

The `for f in (:min, :max, ...) @eval begin...` issue seems similar to coverage not testing for different datatypes (is coverage just the wrong metric?)...

timholy:
@hayd, once this gets merged that change wouldn't be necessary anymore. So I'm not sure it's worth it.

Yes, there absolutely are subtleties (or really, not-so-subtleties...) that this does not handle well. I don't currently have any great ideas about how to take the next step, though. It seems you want _some_ kind of summary number.

jakebolewski:
Getting good coverage results, or any notion of "coverage percentage", is going to be difficult in any language that does runtime code generation.  I think we just have to look at this as a flawed, but useful metric.  Even Stefan's suggestion, instrumenting basic blocks, only will work for code that has actually been generated.  This is complicated by things that generate code at runtime, an example being stagedfunctions.

@timholy, I don't know what packages you have used but maybe we should agree to a non-trivial one to test on.  Glancing at the code, doesn't this have issues with functions that are not at the toplevel of the file.

ex.
```julia
begin
    func() = ...
end
```
or

```julia
module Foo
    module Bar
        func() = .....
    end
end
```

I'll look into making JuliaParser easier to use for this application.  There have been some regressions in terms of compatibility with base during the 0.4 dev period.

timholy:
My earlier worries about non-toplevel functions were erased by later commits---basically I think this is perfect :smile:. Aside from generated code, the one exception I know of is this construction:
```
@doc """
This is the best function I ever wrote.
""" ->
f(x) = 2x
```
The reason is that because of the `->`, the line number for that expression does not appear inside the function body, in contrast with a function defined just as
```
f(x) = 2x
```

timholy:
Oh, regarding your question about testing on a package; my email to julia-dev tested two "mystery" packages, but both have a nontrivial amount of code (>2k lines each, one was over 4k).

timholy:
BTW, I've removed the "RFC"; I'm not planning on pushing any more patches. I do wonder about hooking this into PackageEval and seeing if we can come up with coverage fractions for all registered packages. That could happen either before or after merging this.

IainNZ:
LGTM

IainNZ:
I think getting coverage numbers in PackageEval is definitely possible.

IainNZ:
@timholy can you tag a new version of Coverage when you are ready? I have limited internet access

timholy:
Hmm, I just did a push on Images, and indeed it modified my coverage stats: https://coveralls.io/builds/1701691. (Alternative: https://github.com/timholy/Images.jl/pull/243#issuecomment-68954451) I expect it to go down, but not to 0%. Seems like something is wrong.

The part I can't understand is that the first element in that array of Dicts looks like this:
coverage

which seems entirely reasonable. Any insights about why that file nevertheless is scored as having 0% coverage on Coveralls?

timholy:
Here's a gist of the JSON-encoded output: https://gist.githubusercontent.com/timholy/e5c42bf97adee7ee4333/raw/36edd536817030d278d840976936d11e327504bd/JSON-parsed%20coverage%20stats

timholy:
Oh, actually, Images has had broken coverage data for a long time: the last good one I found dated from August 21, although there must have been a more recent one because the badge is a different number.

https://coveralls.io/r/timholy/Images.jl?page=14

Dunno what's going on, anyone have a clue?

timholy:
OK, I'm getting reports for HDF5, so clearly it's something specific to Images. It would be interesting to debug, but I'm not sure how to go about that.

PR #41, opened 2015-01-07
Modernize .travis.yml, fix the build.

timholy:
It turns out JuliaParser wasn't getting installed because the old script didn't have a Pkg.build("Coverage") in it. So I decided to modernize the `.travis.yml` script, incorporating the recent native support work.

OK if we test on both "nightly" and "release"? (The old one tested only on nightly.) I also noticed that for julia-release, I needed to add Docile to the testing REQUIREments.

Finally, "nightly" won't pass until a new version of JuliaParser gets tagged (https://github.com/jakebolewski/JuliaParser.jl/issues/14). I notice Jake is working hard on the package, and I wouldn't be surprised if he wants to hold off until he finishes his changes.


IainNZ:
Only reason old one only tested on nightly was because `nightly` was Julia 0.3 at the time, needed for coverage.

PR #46, opened 2015-01-21
Add ability to query a repository for its git information

staticfloat:
This supposedly adds a way for `Coverage` to automagically get the git commit information from a repository if it exists, and doesn't hurt anything if it doesn't exist.  I'm having a hard time testing it since I don't think the test submits things to coveralls, but I know at least `Coveralls.query_git_info()` works.

If you want to explicitly disable git information, you call `Coveralls.submit(source_files, git_info_func=nothing)`.  If you have some special way of creating this data, you pass your own callback into `git_info_func`.  Not sure if this is the best API here, but it seems like it should work.

Pinging @timholy for his opinion as well.

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -174,19 +174,65 @@ module Coverage
         #     }
         #   ]
         # }
+
+        import Base.Git
+        function query_git_info(dir="")
+            commit_sha = Git.readchomp(`rev-parse HEAD`, dir=dir)
+            author_name = Git.readchomp(`log -1 --pretty=format:"%aN"`, dir=dir)
+            author_email = Git.readchomp(`log -1 --pretty=format:"%aE"`, dir=dir)
+            committer_name = Git.readchomp(`log -1 --pretty=format:"%cN"`, dir=dir)
+            committer_email = Git.readchomp(`log -1 --pretty=format:"%cE"`, dir=dir)
+            branch = Git.branch(dir=dir)
+            message = Git.readchomp(`log -1 --pretty=format:"%s"`, dir=dir)
+            remote = Git.readchomp(`config --get remote.origin.url`, dir=dir)
+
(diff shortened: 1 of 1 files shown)


IainNZ:
Yeah the tests aren't great. Package design possibly isn't either, didn't really know what it was going to end up looking like (and doing).

I like this PR though! What happens if you're on Travis - is Coveralls going to prefer this info or will it take it from Travis (which is what I'm speculating it does now?)

staticfloat:
I didn't realize it just auto-got it from Travis!  That's bizarrely awesome.  I wonder how that works?

I guess the answer here is that I don't know.  How can we test it?

IainNZ:
We send along the TRAVISID or something like that, which is an environmental variable. It must get everything from that, because Coverage.jl doesn't transfer anything but that.
Its painful, but all I can think of is making a dummy package and turning on Travis/Coveralls, then instead of the usual Coverage code, grab your fork instead...

timholy:
There are some potentially-informative tricks in https://github.com/IainNZ/Coverage.jl/issues/38#issuecomment-69095601.

timholy:
This will be a nice improvement!

staticfloat:
Okay, I think this is ready for review/merging.  I've tested it on `Nettle.jl`, (even though I didn't really need to since I'm modifying `submit_token()` and not `submit()`)

The basic API here is that you pass in a dict or function as `git_info` to `submit_token()`, which then includes that git information in the coverage results if it is able.

IainNZ:
LGTM, @timholy any thoughts?

timholy:
No objections here (I haven't looked at the format this needs to be in, but the code itself looks fine).

IainNZ:
Will tag now

staticfloat:
Thanks, Ian!

PR #50, opened 2015-03-01
Types are hierarchical, which might, just maybe, be important here

staticfloat:
And with this fix, we should be able to get actual commit data on [this page](https://coveralls.io/r/JuliaLang/julia).  It turns out that the dict I was passing in is actually a `Dict{ASCIIString,Any}`, NOT a `Dict`, which makes all the difference.  :P

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -226,9 +226,9 @@ module Coverage
 
             # Attempt to parse git info via git_info, unless the user explicitly disables it by setting git_info to nothing
             try
-                if typeof(git_info) == Function
+                if isa(git_info, Function)
                     data["git"] = git_info()
-                elseif typeof(git_info) == Dict
+                elseif isa(git_info, Dict)
                     data["git"] = git_info
                 end
             end


src/Coverage.jl
```diff
@@ -226,9 +226,9 @@ module Coverage
 
             # Attempt to parse git info via git_info, unless the user explicitly disables it by setting git_info to nothing
             try
-                if typeof(git_info) == Function
+                if typeof(git_info) <: Function
                     data["git"] = git_info()
-                elseif typeof(git_info) == Dict
+                elseif typeof(git_info) <: Dict
                     data["git"] = git_info
```
garborg: `isa(git_info, Dict)`?

IainNZ:
Hah thats brutal. Do you want to make the @garborg change or happy as-is?

IainNZ:
Just gave you commit access too

staticfloat:
Since it looks like you guys know what you're talking about, I'll make the change.  What benefit does `isa(x, foo)` give over `typeof(x) <: foo`?  Is it just aesthetic?

garborg:
Just aesthetic :) Not worth holding holding up your fix, or course, but when functionality is exported from Base and the choice is between using it and writing out its definition, I'm a fan of embracing either embracing it or lobbying for its removal.

staticfloat:
Yeah!  [Good work everybody](https://coveralls.io/r/JuliaLang/julia).

timholy:
Nice!

PR #68, opened 2015-07-06
RFC: Modified Coverage to work with .{pid}.cov format

kshyatt:
I've tested this on Coveralls and Codecov. I have a PR I am about to submit to main Julia which will modify things so that each process writes `{filename}.{pid}.cov` instead of `{filename}.cov` as before. I've written some functions to handle the new format. I've checked also that my functions work with the old format and everything seems fine. `merge_coverage_counts` was lifted from CoverageBase - hopefully @timholy is ok with that.

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -9,14 +9,26 @@ module Coverage
     using Compat
 
     export process_cov, amend_coverage_from_src!
+    export process_cov_with_pids
     export coverage_file, coverage_folder
-    export analyze_malloc
+    export analyze_malloc, merge_coverage_counts
 
+    function merge_coverage_counts(a1, a2)
+        n = max(length(a1),length(a2))
+        a = Array(Union(Void,Int), n)
+        for i = 1:n
+            a1v = isdefined(a1, i) ? a1[i] : nothing
+            a2v = isdefined(a2, i) ? a2[i] : nothing
+            a[i] = a1v == nothing ? a2v :
+                   a2v == nothing ? a1v : max(a1v, a2v)
+        end
+        a
+    end
     # process_cov
     # Convert a Julia .cov file into an array of (counts, lines)
     #
     # Input:
-    # filename          Coverage file to open
+    # filename          Coverage files to open
     #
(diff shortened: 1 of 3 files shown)


kshyatt:
!!! I forgot to update the internal tests to work with my newly changed functions. Will rebase!

kshyatt:
I'm dying on `0.3`, but passing on `0.4`, which makes sense because I only have `0.4` running at home.

timholy:
I approve of the pick-pocketing :smile:. In retrospect putting the merge functionality here, rather than in CoverageBase, makes more sense. Once this gets merged, feel free to delete from CoverageBase.

timholy:
You could bump the REQUIRE file. I think the travis test will test this PR even if it's contrary to what REQUIRE specifies. @IainNZ would have to comment, though, about how to handle that. Probably best to make a release-0.3 branch or something, in case further improvements for 0.3 need to be made.

kshyatt:
So the tests are dying on the new code which would be 0.4 only anyway. I guess I could write a macro to only run the pid tests if we're on julia-nightly rather than julia-release. What does everyone think?

IainNZ:
Yeah can't there just be a 0.4-only-if in the tests?

kshyatt:
Yes, I think I have it now. We'll see...

kshyatt:
Is this new failure in `nightly` my fault? If not, I think we are good to merge.

IainNZ:
It was working yesterday: https://travis-ci.org/IainNZ/Coverage.jl/builds/69546901
Its weird, Travis partially passed. Is this branch based off of master HEAD? Like `git rebase master` is a no-op?

IainNZ:
Oh it just failed to download Julia itself: 

```
The command "curl -s -L --retry 7 'https://status.julialang.org/download/linux-x86_64' | tar -C ~/julia -x -z --strip-components=1 -f -" failed and exited with 2 during .
```

IainNZ:
I think its fine then

kshyatt:
Here goes nothing...

PR #76, opened 2015-07-24
Big Refactor

IainNZ:


IainNZ:
OK @staticfloat @kshyatt its done. This is a breaking change if you were using the internals, as I'm guessing the buildbots might do? I can fix that code if you point me to it, so I can merge this.

kshyatt:
@IainNZ I can take a look at CoverageBase tonight, that's what will need to be changed. @timholy might want to take a look too. Ideally we can leave the buildbots mostly untouched...

IainNZ:
OK, cool. Conceptually, biggest change is process_file returns a FileCoverage object, and process_folder returns a vector of these. I'm expecting most breakage to come from that.

timholy:
Looks like a nice cleanup to me, thanks for doing this.

PR #79, opened 2015-08-15
Switched from JuliaParser to Base

kshyatt:
This works now with the newest `Base` master.


IainNZ:
Looks like it might need a `readall` on 0.3, and a test result change on 0.4.

kshyatt:
Got the `0.4` change done. Where is the `readall` needed?

IainNZ:
There error message on 0.3 sez `ERROR: `parse` has no method matching parse(::IOStream)`

IainNZ:
I guess it only takes strings?

kshyatt:
I talked with @IainNZ and we're going to merge then get set for 0.4, which should be coming very soon.

PR #82, opened 2015-08-20
Emit helpful warning error instead of BoundsError

timholy:
Helps idiots like me.


IainNZ:
Need. While you're in the vicinity, any ideas about https://github.com/IainNZ/Coverage.jl/issues/81?

timholy:
Not without some digging. Will have to be later---got Colors to fry!

IainNZ:
:+1:

PR #83, opened 2015-08-24
Fix #81

### LINKED ISSUE #81: `module X` incorrectly marked as not covered?
Something different about how Base.parse works, perhaps

simonster:
The problem is the `eval` definitions here, which have corresponding line number nodes since JuliaLang/julia@cb2dbda6e00b202c739b1bc0c068616878744984:

```julia
julia> :(module X end)
:(module X
    eval(x) = begin  # none, line 1:
            top(Core).eval(X,x)
        end
    eval(m,x) = begin  # none, line 1:
            top(Core).eval(m,x)
        end
    end)
```

Those `eval` definitons may not necessarily get hit, but probably shouldn't count against coverage.


kshyatt:
LGTM!

PR #85, opened 2015-08-24
Add SimJulia to the packages using Coverage

BenLauwens:
Thanks for this great package!

Ben

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ after_success:
 * [RDF.jl](https://github.com/joejimbo/RDF.jl/blob/master/.travis.yml)
 * [Requests.jl](https://github.com/loladiro/Requests.jl/blob/master/.travis.yml)
 * [Robotics.jl](https://github.com/cdsousa/Robotics.jl/blob/master/.travis.yml)
+* [SimJulia.jl](https://github.com/BenLauwens/SimJulia.jl/blob/master/.travis.yml)
 * [SIUnits.jl](https://github.com/loladiro/SIUnits.jl/blob/master/.travis.yml)
 * [StatsBase.jl](https://github.com/JuliaStats/StatsBase.jl/blob/master/.travis.yml)
 * [TextWrap.jl](https://github.com/carlobaldassi/TextWrap.jl/blob/master/.travis.yml)


IainNZ:
Thanks for SimJulia! We taught a ~1 hour class using it at MIT, and possibly will teach a longer one again in January.

BenLauwens:
Hi Iain

Nice to hear that it is used at MIT.
This semester I will use SimJulia for two courses:
- reliability of systems 
- performance analysis of telecommunication systems

At the RMA (Royal Military Academy) we are using it actively in the following research areas:
- tactical radio communication systems
- survivability of combat vehicles and their crew
- (military) human resource planning
- medical battlefield management
- maritime demining planning/optimisation

If there is some interest I can always give a talk at the next JuliaCon on SimJulia internals or its applications.
Kind regards

Ben

BenLauwens:
Hi Iain

Can you tell me something more about the class you already gave using SimJulia and the longer one in January?
The development of SimJulia was done in my spare time. We are funded based on applied research projects and developing a simulation framework is not a primary goal. I like to show that SimJulia is more than a toy project and has broad applications.
Thanks

Ben

IainNZ:
Hi Ben,

At MIT, there are no normal classes in January. Instead we have an "Independent Activities Period", that contains shorter special-interest courses (up to 4 weeks long). We (doctoral candidates in my Operations Research program) created a short course to teach software tools to our classmates. We actually wrote a paper after the second iteration of the class (January 2014):
http://pubsonline.informs.org/doi/abs/10.1287/ited.2014.0131

In January 2015 one of the volunteers wanted to teach simulation using SimPy. Beacuse we wanted to teach the class only in Julia and R though, they ran the class (about 1.5 hours) in Julia with SimJulia instead. I think they had a bit of trouble because they were used to SimPy 3 but at the time SimJulia was more like SimPy 2? But its was reasonably popular, and it was the first taste of this type of software for a lot of students.

I'm not aware of anyone using it for research right now in my department, but we don't do much simulation work here in general.

Thanks,
Iain

BenLauwens:
Hi Iain

Thanks. SimJulia v0.3+ has the same API as SimPy v3. So the migration from SimPy will be easier this time.
Kind regards

Ben

PR #105, opened 2016-05-30
Generic codecov submissions to support enterprise installations

ccoffrin:
I am fairly new to Julia so I might have done this all wrong, please don't hesitate to tell me so.  :-)

I ran into a problem with the current Codecov implementation where I could not submit reports to an enterprise installation of Codecov.  The key problem was that the uri needs to point to something like "enterprise-codecov.com" instead of "codecov.io", and even if I am using Travis-CI I need to provide a repo token.

Summary of the proposed changes:
- adding a new generic submission method (i.e. `generic_submit`)
- re-casting existing submission methods in terms of `generic_submit`
- deprecating `submit_token` in favor of a hopefully more intuitive name `submit_local`
- making environment variables consistent with [codecov-bash](https://github.com/codecov/codecov-bash) 
- deprecating `REPO_TOKEN` in favor of `CODECOV_TOKEN` used by codecov-bash

The approach:
To fix my enterprise codecov issues, and hopefully guard against similar issues in the future, I took the approach of making a generic_submit function, which builds the codecov uri from the function's keyword arguments.  Hence, the upload uri can be adapted or extended easily by passing new keyword arguments this function.

With the new generic_submit, we can now see submit and submit_token as simply defining some default values to pass to generic_submit.  Any of these default values can be overridden simply by passing the appropriate keyword argument to those functions.  When no default value or keyword argument are specified, environment variables are checked and used if available.

I have verified this works on enterprise use cases, but have not tested the non-enterprise uses cases.

### DIFF
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -105,7 +105,6 @@ module Codecov
     @deprecate submit_token submit_local
 
 
-
     """
         submit_generic(fcs::Vector{FileCoverage})
 
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -48,7 +48,7 @@ module Codecov
     function set_defaults(args_array; kwargs...)
         defined_names = [k for (k,v) in args_array]
         for kwarg in kwargs
-            if ~(kwarg[1] in defined_names)
+            if !(kwarg[1] in defined_names)
                 push!(args_array, kwarg)
             end
         end
@@ -81,13 +81,17 @@ module Codecov
     """
         submit_token(fcs::Vector{FileCoverage})
 
-    make depreciated
+    Takes a `Vector` of file coverage results (produced by `process_folder`),
+    and submits them to Codecov.io. Assumes the submission is being made from 
+    a local git installation.  A repository token should be specified by a 
+    'token' keyword argument or the CODECOV_TOKEN environment variable.
     """
-    function submit_token(fcs::Vector{FileCoverage})
+    function submit_token(fcs::Vector{FileCoverage}; kwargs...)
         println("submit_token is deprecated, use submit_local instead")
-        submit_local(fcs)
+        submit_local(fcs; kwargs...)
     end
 
+    @deprecate submit_token submit_local
 
     import Base.Git
 
@@ -170,3 +174,4 @@ module Codecov
     end
 
 end  # module Codecov
+
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -114,7 +114,7 @@ module Codecov
     values match the Codecov upload/v2 API specification.  
     The `codecov_url` keyword argument or the CODECOV_URL environment variable 
     can be used to specify the base path of the uri.
-    The `julia_test` keyword can be used to prevent the http request from 
+    The `dry_run` keyword can be used to prevent the http request from 
     being generated
     """
     function submit_generic(fcs::Vector{FileCoverage}; kwargs...)
@@ -129,20 +129,20 @@ module Codecov
         end
 
         codecov_url = "https://codecov.io"
-        julia_test = false
+        dry_run = false
         for (k,v) in kwargs
             if k == :codecov_url
                 codecov_url = v
             end
-            if k == :julia_test
-                julia_test = true
+            if k == :dry_run
+                dry_run = true
             end
         end
         @assert codecov_url[end] != "/" "the codecov_url should not end with a /, given url $(codecov_url)"
 
         uri_str = "$(codecov_url)/upload/v2?"
         for (k,v) in kwargs
-            if k != :codecov_url && k != :julia_test
+            if k != :codecov_url && k != :dry_run
                 uri_str = "$(uri_str)&$(k)=$(v)"
             end
         end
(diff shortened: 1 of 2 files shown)


src/codecovio.jl
```diff
@@ -40,53 +40,129 @@ module Codecov
...
+
+        codecov_url = "None"
+        for (k,v) in kwargs
+            if k == :codecov_url
+                codecov_url = v
+                break
+            end
+        end
+        @assert codecov_url != "None"
+        @assert codecov_url[end] != "/"
```
tkelman: should these give better error messages if people use this incorrectly?

src/codecovio.jl
```diff
@@ -40,53 +40,129 @@ module Codecov
...
 
+
+    """
+    kwargs provides default values to insert into args_array, only if they are 
+    not already specified in args_array.
+    """
+    function set_defaults(args_array; kwargs...)
+        defined_names = [k for (k,v) in args_array]
+        for kwarg in kwargs
+            if ~(kwarg[1] in defined_names)
```
tkelman: `~` is bitwise, usually `!` is preferred for boolean not

src/codecovio.jl
```diff
@@ -40,53 +40,129 @@ module Codecov
...
+        submit_generic(fcs; kwargs...)
     end
 
+
+    """
+        submit_token(fcs::Vector{FileCoverage})
+
+    make depreciated
+    """
+    function submit_token(fcs::Vector{FileCoverage})
```
tkelman: there's an `@deprecate` macro you can use

ccoffrin:
In these commits I,
- addressed the minor suggestions of @tkelman 
- made the default codecov_url value easier to override
- added tests to ensure the urls are being generated correctly

I not 100% sure I used the @depricated macro correctly, @tkelman can you have a look at line 94?

src/codecovio.jl
```diff
@@ -40,53 +40,138 @@ module Codecov
...
+    and submits them to Codecov.io. Assumes the submission is being made from 
+    a local git installation.  A repository token should be specified by a 
+    'token' keyword argument or the CODECOV_TOKEN environment variable.
+    """
+    function submit_token(fcs::Vector{FileCoverage}; kwargs...)
+        println("submit_token is deprecated, use submit_local instead")
+        submit_local(fcs; kwargs...)
+    end
+
+    @deprecate submit_token submit_local
```
tkelman: using the `@deprecate` macro should allow you to leave out the placeholder definition of `submit_token` above

test/runtests.jl
```diff
@@ -61,3 +61,196 @@ cd(Pkg.dir("Coverage")) do
...
+    #println("url: $(url)")
+    @assert url != "None" "unable to find codecov api url in stdout, check for changes in codecovio.jl" 
+    return url
+end
+
+
+# clear system defined ENV vars in preparation for testing
+
+delete!(ENV, "CODECOV_URL")
+delete!(ENV, "CODECOV_TOKEN")
```
tkelman: would probably be better to use `withenv` for temporarily clearing or setting these and automatically restoring them

src/codecovio.jl
```diff
@@ -40,53 +40,138 @@ module Codecov
...
+    """
+        submit_generic(fcs::Vector{FileCoverage})
+
+    Takes a vector of file coverage results (produced by `process_folder`),
+    and submits them to a Codecov.io instance. Keyword arguments are converted 
+    into a generic Codecov.io API uri.  It is essential that the keywords and 
+    values match the Codecov upload/v2 API specification.  
+    The `codecov_url` keyword argument or the CODECOV_URL environment variable 
+    can be used to specify the base path of the uri.
+    The `julia_test` keyword can be used to prevent the http request from
```
tkelman: should this be called something different? `dry_run` maybe?

ccoffrin:
In these commits I integrated the suggestions of @tkelman.  @depricated now seems to be working as expected.

test/runtests.jl
```diff
@@ -61,3 +61,193 @@ cd(Pkg.dir("Coverage")) do
...
+
+    # default values
+    codecov_url = extract_codecov_url( () -> Coverage.Codecov.submit_local(fcs; dry_run = true) )
+    @test contains(codecov_url, "codecov.io")
+    @test contains(codecov_url, "commit")
+    @test contains(codecov_url, "branch")
+    @test !contains(codecov_url, "service")
+
+    # default values in depreciated call
+    codecov_url = extract_codecov_url( () -> Coverage.Codecov.submit_token(fcs) )
```
ccoffrin: @tkelman is there a standard/simple way that tests can check for a warning?

This [thread](https://groups.google.com/forum/#!msg/julia-users/X90Tvh4kixU/DNbB-wJFacAJ) suggests there is not a simple way to do it.

ccoffrin:
Are there any more concerns with these changes?

timholy:
On a brief scan this looks fine to me: it seems very thorough, and it passes tests on julia-0.4 (on 0.5, there's an unrelated error). Since you put some time into reviewing this @tkelman, I'll ask: OK if I merge this?

ccoffrin:
Thanks a bunch!

PR #108, opened 2016-07-13
Prepare for upcoming deletion of Base.Git module

tkelman:
(to whom it may concern...)

https://github.com/JuliaPackaging/Git.jl should be essentially equivalent, but will also ensure that you have git installed on platforms where it knows how to get a binary. If you already have git on your system path it'll just use it.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,3 +1,4 @@
 julia 0.4
 JSON
 Requests
+Git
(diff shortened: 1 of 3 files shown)


tkelman:
According to http://pkg.julialang.org/detail/Coverage.html the nightly failure started sometime in the pkgeval gap between march and may.

PR #111, opened 2016-07-18
print result of codecov submission in utf-8 rather than UInt8

vtjnash:


src/codecovio.jl
```diff
@@ -155,7 +155,7 @@ module Codecov
             data    = to_json(fcs)
             req     = Requests.post(URI(uri_str); json = data, headers = heads)
             println("Result of submission:")
-            dump(req.data)
+            println(String(req.data))
```
tkelman: needs Compat for 0.4?

PR #112, opened 2016-07-18
print result of coveralls submission in utf-8 rather than UInt8

vtjnash:


tkelman:
I'm a little surprised this passed but #111 didn't...

vtjnash:
Missing test coverage?

tkelman:
likely. can you add Compat (with minimum version requirement) to be on the safe side? and more tests if you know how.

PR #113, opened 2016-07-20
hide token from Codecov.io submission URL logs

vtjnash:


vtjnash:
hm. it seems the tests are depending on the token being printed to the log. thoughts?

tkelman:
fix the tests? in what way do they depend on that?

vtjnash:
they assert that we will print the token into the logs (e.g. https://github.com/JuliaCI/Coverage.jl/blob/master/test/runtests.jl#L227) :/

i guess we could print the token if the user passed `dry_run`?

tkelman:
either that, or remove that line

tkelman:
bump

fingolfin:
Would be kinda nice to resurrect this PR

fingolfin:
This was reimplemented in PR #222

PR #127, opened 2017-05-12
Depend and use HTTP instead of Requests and drop 0.4

musm:
btw the tests don't really work on windows
one of them even causes julia to segfault... 

I will have to try on a linux machine to try and be able to get the test script running so I can debug


musm:
Closing, due to lack of interest by maintainers

PR #129, opened 2017-05-27
fix include bug, escape_string test filename

musm:
This line fails since it is not properly escaped  in windows where the  path delimiter is `\`


musm:
cc @tkelman

musm:
bump

musm:
bump

musm:
bump

musm:
anyone?

tkelman:
we should probably turn appveyor on here

Issue #98, opened 2015-11-27
Document calling from command line


jgoldfar:
I would submit a PR if I knew exactly what the intended use was, but this https://github.com/IainNZ/Coverage.jl/blob/master/src/memalloc.jl#L58 caught me when I was trying to write a coverage command line script.

IainNZ:
ping @timholy, who wrote that

timholy:
That line contains everything I remember about this (and more). Looks like it has to be run from `Coverage/src`, and you'd presumably have to update the `find` path to 0.4.

Do you have specific questions?

spk83:
@timholy I ran into issue because of the line pointed by @jgoldfar . I am using Coverage.jl in a command line script and I was passing arg to the script which was being read by the https://github.com/JuliaCI/Coverage.jl/blob/master/src/memalloc.jl#L60 as a file, very surprising. I don't understand why it has to read args from command line. Because of this you cant use Coverage.jl in a script which you want to run with some args in non-interactive mode

omus:
I think we should delete the [non-interactive logic in Coverage.jl](https://github.com/JuliaCI/Coverage.jl/blob/08f64bcc43ccac1b10382383db40942997c6669f/src/memalloc.jl#L58-L62) as it is an issue for scripts which import Coverage and it's rather trivial to do with:

```
julia -e "using Coverage; println(Coverage.analyze_malloc_files(ARGS))"
```

jgoldfar:
It would probably be prudent to find a way to estimate how much that would break, seeing as it was added for a reason at some point.

PR #131, opened 2017-06-30
Let programs using Coverage use command-line arguments

christopher-dG:
This is weird to explain so here's an example:

```
[ ~ ] cat > x.jl
using Coverage
print(ARGS)
[ ~ ] julia x.jl fjsf
ERROR: LoadError: LoadError: LoadError: SystemError: opening file fjsf: No such file or directory
Stacktrace:
 [1] #systemerror#37 at ./error.jl:64 [inlined]
 [2] systemerror(::String, ::Bool) at ./error.jl:64
 [3] open(::String, ::Bool, ::Bool, ::Bool, ::Bool, ::Bool) at ./iostream.jl:104
 [4] open(::Coverage.##17#18{Array{Coverage.MallocInfo,1}}, ::String) at ./iostream.jl:150
 [5] analyze_malloc_files(::Array{String,1}) at /Users/degraafc/.julia/v0.6/Coverage/src/memalloc.jl:20
 [6] include_from_node1(::String) at ./loading.jl:539
 [7] include(::String) at ./sysimg.jl:14
 [8] include_from_node1(::String) at ./loading.jl:539
while loading /Users/degraafc/.julia/v0.6/Coverage/src/memalloc.jl, in expression starting on line 60
while loading /Users/degraafc/.julia/v0.6/Coverage/src/Coverage.jl, in expression starting on line 272
while loading /Users/degraafc/x.jl, in expression starting on line 1
```

`isinteractive()` only deals with the REPL case, not the `using` case. This change deals with both! It looks like this problem has previously been identified in https://github.com/JuliaCI/Coverage.jl/issues/98#issuecomment-218820915.

### DIFF
--- a/src/memalloc.jl
+++ b/src/memalloc.jl
@@ -56,7 +56,7 @@ isfuncexpr(ex::Expr) =
 isfuncexpr(arg) = false
 
 # Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
-if abspath(PROGRAM_FILE) == Pkg.dir(joinpath("Coverage", "src", "Coverage.jl"))
+if abspath(PROGRAM_FILE) == joinpath(dirname(@__FILE__), "Coverage.jl")
     bc = analyze_malloc_files(ARGS)
     println(bc)
 end
--- a/src/memalloc.jl
+++ b/src/memalloc.jl
@@ -56,7 +56,7 @@ isfuncexpr(ex::Expr) =
 isfuncexpr(arg) = false
 
 # Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
-if abspath(PROGRAM_FILE) == joinpath(dirname(@__FILE__), "Coverage.jl")
+if abspath(PROGRAM_FILE) == joinpath(@__DIR__, "Coverage.jl")
     bc = analyze_malloc_files(ARGS)
     println(bc)
 end
--- a/src/memalloc.jl
+++ b/src/memalloc.jl
@@ -56,7 +56,7 @@ isfuncexpr(ex::Expr) =
 isfuncexpr(arg) = false
 
 # Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.3 -name "*.jl.mem")`
-if !isinteractive()
+if abspath(PROGRAM_FILE) == @__FILE__
     bc = analyze_malloc_files(ARGS)
     println(bc)
 end
--- a/src/memalloc.jl
+++ b/src/memalloc.jl
@@ -55,8 +55,8 @@ isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
 
-# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.3 -name "*.jl.mem")`
-if abspath(PROGRAM_FILE) == @__FILE__
+# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
+if abspath(PROGRAM_FILE) == Pkg.dir(joinpath("Coverage", "src", "Coverage.jl"))
     bc = analyze_malloc_files(ARGS)
     println(bc)
 end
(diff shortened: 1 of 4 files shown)


christopher-dG:
Note: This won't fix 0.4 because `PROGRAM_FILE` isn't defined.

tkelman:
dropping 0.4 support is fine

christopher-dG:
Cool, updated.

tkelman:
any straightforward way to test this?

christopher-dG:
I'm assuming you mean in a CI environment.

The only way I can think of is putting `using Coverage` in a file, and then running that with some command-line arguments just to make sure it doesn't break. So something along the lines of:

```yaml
script:
  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
  - julia --check-bounds=yes -e 'Pkg.clone(pwd()); Pkg.test("Coverage"; coverage=true)'
  - echo "using Coverage" > test/using.jl"  # Could just add this file to the repo
  - julia test/using.jl gibberish
```

Not very pretty, but it more or less does the job.

omus:
@christopher-dG it would be better to included this as a test that is run during `Pkg.test`. 
For validating that the change is working you can do:
```julia
script = tempname()
write(script, """
    using Coverage; println(join(ARGS, ","))
    """)
result = try
    readchomp(`$(Base.julia_cmd()) $script argument`)
catch
   ""
end
@test result == "argument"
```
You should do another test for ensuring that you can still use the non-interactive code.

src/memalloc.jl
```diff
@@ -55,8 +55,8 @@ isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
 
-# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.3 -name "*.jl.mem")`
-if abspath(PROGRAM_FILE) == @__FILE__
+# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
+if abspath(PROGRAM_FILE) == Pkg.dir(joinpath("Coverage", "src", "Coverage.jl"))
```
tkelman: don't use Pkg.dir. if `PROGRAM_FILE` doesn't match `@__FILE__` but something adjacent to it, use `joinpath(dirname(@__FILE__), "Coverage.jl")`

test/runtests.jl
```diff
@@ -61,6 +61,21 @@ cd(dirname(@__DIR__)) do
     open("fakefile",true,true,true,false,false)
     @test isempty(Coverage.process_cov("fakefile",datadir))
     rm("fakefile")
+
+    # Test `using Coverage` with non-empty command-line arguments
+    script = tempname()
```
tkelman: delete the `script` when you're done with it

src/memalloc.jl
christopher-dG: Didn't think of that, thanks.

test/runtests.jl
christopher-dG: Is this necessary when the file is in a temp directory?

tkelman: some operating systems clean up temporary directories automatically, but better not to have to rely on that

src/memalloc.jl
```diff
@@ -55,8 +55,8 @@ isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
 
-# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.3 -name "*.jl.mem")`
-if !isinteractive()
+# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
+if abspath(PROGRAM_FILE) == joinpath(dirname(@__FILE__), "Coverage.jl")
```
omus: You can use `@__DIR__` instead of `dirname(@__FILE__)`. This should work on all versions of Julia since Compat is already requiring Compat 0.9.5.

src/memalloc.jl
```diff
@@ -55,8 +55,8 @@ isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
 
-# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.3 -name "*.jl.mem")`
-if !isinteractive()
+# Support Unix command line usage like `julia Coverage.jl $(find ~/.julia/v0.6 -name "*.jl.mem")`
+if abspath(PROGRAM_FILE) == joinpath(@__DIR__), "Coverage.jl")
```
omus: Extra parentheses after `@__DIR__`

PR #133, opened 2017-08-24
Add support for Circle CI

iamed2:
It's the third CI platform supported by/supporting CodeCov by default so I think it's reasonable to support it in Coverage.jl.

I didn't include the `job` key (there's no unique job number in Circle CI) but I did include the `build_url` key.


iamed2:
Nightly failure is identical to master's; tests are passing on the other Julia versions.

iamed2:
You're welcome! Thanks for reviewing!

Once I get Circle CI working for a test package I'm going to try to add it to PkgTemplates.jl and maybe also `PkgDev.generate`.

PR #134, opened 2017-08-24
Fix Circle CI PR number not available on non-PR builds

iamed2:
Apparently this variable is unset on non-PR builds. Travis sets it to `"false"` in those cases, so that should be fine here.


ararslan:
Nightly failure appears unrelated.

PR #135, opened 2017-09-02
typo: else instead of elseif

gaika:
Breaks on symbolic links

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -207,7 +207,7 @@ module Coverage
                 else
                     println("Coverage.process_folder: Skipping $file, not a .jl file")
                 end
-            else isdir(fullfile)
+            elseif isdir(fullfile)
                 # If it is a folder, recursively traverse
                 append!(source_files, process_folder(fullfile))
             end


tkelman:
how did that work at all? would be good to add a test that hits this branch

musm:
bump @tkelman @ararslan

ararslan:
Still needs a test that hits that branch

musm:
true, @gaika  want to add a test ? (seems  unlikely, since I think this was a regex search of repos incorrectly using else instead of elseif), otherwise it's better to merge this and have the bug fixed without a test rather than keeping than a bug and no test.

gaika:
I don't know how to add a test for it, in my case it only fails when there are emacs backup files (symbolic links to a pipe?) in the directory to cover.

ararslan:
You could try adding a file like that in the directory that's processed as part of the tests.

gaika:
For some reason It doesn't trigger with the test data in the Coverage.jl, here's a stack trace from my test repo:
```
ERROR: SystemError: unable to read directory /home/joe/.julia/v0.6/Test/src/.#
Test.jl: No such file or directory                                           
Stacktrace:
 [1] readdir(::String) at ./file.jl:415
 [2] process_folder(::String) at /home/joe/.julia/v0.6/Coverage/src/Coverage.jl:200                                                                        
 [3] process_folder(::String) at /home/joe/.julia/v0.6/Coverage/src/Coverage.jl:212   
```

Where `.#Test.jl -> joe@test.com.6223:1509759031`

ararslan:
Hm, I'm not really sure what to make of that error.

gaika:
Happens with `Coverage.process_folder(dir)` when there are files still being edited (and locked) in Emacs. The lock file is not a file, so it hits the else clause (instead of elseif) and assumes it is a directory and tries to process it recursively

ararslan:
Okay. Well, what you've change this to is clearly what was intended to be written. We can always figure out an appropriate test for this later, but in the meantime we might as well merge this as-is.

PR #136, opened 2017-09-02
Fix tests on Julia v0.7

kmsquire:
* Parsing seems to have changed slightly in Julia (or JuliaParser?) between v0.6 and v0.7.
  For test/data/testparser.jl, line 10 is the end of a @doc string (`""" ->`), and Line 11
  is an expression (`f6(x) = 6x`)
* In v0.6, the zero count goes to line 10, and in v0.7, the zero count goes (more correctly?)
  to line 11

PR #137, opened 2017-10-24
Fix tests on 0.6.1

ararslan:
The commit that broke the tests here on 0.7 (which was fixed by #136) was backported to 0.6.1. Once that's released, the tests will fail on the latest 0.6 release, so this PR adjusts the test to look for all applicable versions.


ararslan:
Nightly failure is unrelated.

PR #145, opened 2018-01-23
Get rid of the dependence on Git.jl

kshyatt:


src/codecovio.jl
```diff
@@ -143,15 +143,15 @@ module Codecov
...
-    a local git installation.  A repository token should be specified by a
-    'token' keyword argument or the `CODECOV_TOKEN` environment variable.
+    Take a `Vector` of file coverage results (produced by `process_folder`),
+    and submit them to Codecov.io. Assumes the submission is being made from
+    a local git installation, rooted at `dir`. A repository token should be specified by a
+    `token` keyword argument or the `CODECOV_TOKEN` environment variable.
     """
-    function submit_local(fcs::Vector{FileCoverage}; kwargs...)
-        LibGit2.with(LibGit2.GitRepo(pwd())) do repo
+    function submit_local(fcs::Vector{FileCoverage}, dir::AbstractString=pwd(); kwargs...)
```
ararslan: Should `joinpath(pwd(), "..")` be the default if that's what's needed to get the tests to work?

kshyatt: Does `Pkg.test` run things in `PKG_ROOT/test` or `PKG_ROOT`?

ararslan: I could be mistaken, but looking at the code it appears to run in `PKG_ROOT/test`.

src/codecovio.jl
```diff
@@ -140,21 +140,27 @@ module Codecov
...
-            branch = Git.branch(dir="")
-            )
+    function submit_local(fcs::Vector{FileCoverage}, dir::AbstractString=joinpath(pwd(),".."); kwargs...)
+        LibGit2.with(LibGit2.GitRepo(dir)) do repo
+            LibGit2.with(LibGit2.head(repo)) do headref
+                branch_name = LibGit2.shortname(headref)
+                commit_oid  = LibGit2.GitHash(LibGit2.peel(headref))
+                kwargs = set_defaults(kwargs,
+                    commit = string(commit_oid),
+                    branch = LibGit2.branch(headref) # this function returns a String
```
ararslan: This appears to be triggering the error in the PR build: `ERROR: LoadError: GitError(Code:ERROR, Class:Invalid, Reference 'HEAD' is neither a local nor a remote branch.)` I wonder if it's because Travis checks out the merge commit for PR builds.

kshyatt: Interesting...

ararslan: I could be wrong, but according to the documentation for the LibGit2 module, `LibGit2.branch` _creates_ a new branch. If you look at the definition of `Git.branch` from the Git package, which was used previously, it appears to be getting the name of the branch currently being built: https://github.com/JuliaPackaging/Git.jl/blob/master/src/Git.jl#L165.

kshyatt: yeaaaaaah oh no i did this wrong will fix

kshyatt: WAIT

src/codecovio.jl
```diff
@@ -140,21 +140,27 @@ module Codecov
...
-            branch = Git.branch(dir="")
-            )
+    function submit_local(fcs::Vector{FileCoverage}, dir::AbstractString=joinpath(pwd(),".."); kwargs...)
+        LibGit2.with(LibGit2.GitRepo(dir)) do repo
+            LibGit2.with(LibGit2.head(repo)) do headref
+                branch_name = LibGit2.shortname(headref)
+                commit_oid  = LibGit2.GitHash(LibGit2.peel(headref))
+                kwargs = set_defaults(kwargs,
+                    commit = string(commit_oid),
+                    branch = LibGit2.shortname(headref) # this function returns a String
```
ararslan: Couldn't you use `branch_name` here, which you computed above?

ararslan: Or just remove that variable definition in favor of using this inline, since it otherwise isn't used.

ararslan:
0.7 failures are unrelated.

PR #148, opened 2018-01-23
Restructure Travis tests to be more transparent

ararslan:
Currently coverage information is submitted in the `after_success` block of the Travis script, as it is in most packages. This makes sense for the vast majority of packages, since if coverage submission fails you don't want it to fail your build. However, that's exactly what we want to happen if submission fails here.

This commit restructures the tests to put the coverage submission in the `script` block. This should cause the build to fail if submission fails.


kshyatt:
closed/reopened to get CI to try to run against @quinnj's fix

kshyatt:
failure unrelated, im gonna merge

PR #149, opened 2018-01-23
Aa/coveralls

quinnj:


kshyatt:
Is this the one I want to merge?

kshyatt:
Can be done later but we should really change all those printlns to use the base logging API

src/coveralls.jl
```diff
@@ -11,6 +11,7 @@ module Coveralls
     using HTTP
     using JSON
     using Compat
+    using MbedTLS
```
tkelman: this needs to be in require, don't import packages that you don't directly depend on

kshyatt: Good catch, sorry

PR #152, opened 2018-01-24
Try to see if the problem is in Codecov or process_folder

kshyatt:


kshyatt:
It's happening for both Coveralls and Codecov ... looks like a `process_cov` issue to me? Could it be that when we `Pkg.rm` we're deleting all the `.cov` files?

PR #157, opened 2018-01-25
Use the newest HTTP to see what happens

ararslan:
Mostly opening this to see what CI says. If there are changes in HTTP that break Coverage then I'll use this PR to fix them.

### DIFF
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,5 +1,5 @@
 julia 0.6
 JSON
-HTTP
+HTTP 0.6.0
 MbedTLS
 Compat 0.44.0


kshyatt:
Looking good! 🎉

fingolfin:
What's the status of this? Perhaps rebase it?

fingolfin:
@ararslan ping? This will conflict with #191.

ararslan:
> This will conflict with #191.

I don't see why that would be the case, but at any rate, this PR is not necessary. Pkg will pick up the newest compatible version of HTTP, which has exceeded 0.6.0.

fingolfin:
This PR modifies REQUIRE, the other PR deletes it => conflict (of course one that's trivial to resolve)

PR #158, opened 2018-02-08
Fix parser positioning issue

kshyatt:
Checked against julia master, this works.


kshyatt:
lights are green let's roll

PR #161, opened 2018-04-19
Update for the most recent Meta.parse

gaika:
`Meta.parse` API and order of parameter evaluation has changed. Not sure about exact Julia version for the changes.


fingolfin:
This patch is certainly needed for the released versions of Julia 0.7 and 1.0.

Moreover, I don't think the code before this patch *ever* worked. I'll submit an issue, and hope that this PR will be merged soon (but then it's been not merged nor reviewed for 6 months... :-/ )

PR #165, opened 2018-07-08
Support for codecov from Jenkins

laurentheirendt:
cc: @syarra


laurentheirendt:
cc: @ararslan

ararslan:
I'm not familiar with Jenkins so as long as you've verified that it works for you on Jenkins then this is fine with me.

laurentheirendt:
Yes, the relevant code lines are tested and are currently in use on [ARTENOLIS](https://opencobra.github.io/artenolis). The respective jobs are [here](http://artenolis.lcsb.uni.lu/).

PR #164, opened 2018-07-05
Adding verbose flag to codecovio

laurentheirendt:
This addition of the `verbose` flag avoids printing the token on a CI when submitting the coverage report.

PR #166, opened 2018-07-12
Adding verbose mode and support for coveralls from Jenkins

laurentheirendt:
This PR combines PR https://github.com/JuliaCI/Coverage.jl/pull/164 and the request from @ararslan to implement a verbose mode for coveralls.

Similar to PR https://github.com/JuliaCI/Coverage.jl/pull/165, I added the support for coveralls from Jenkins.

cc: @syarra

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -101,7 +101,10 @@ module Coveralls
 
             data["git"] = query_git_info()
 
-            data["git"]["branch"] = split(ENV["GIT_BRANCH"], "/")[2]
+            # get the name of the branch if not a pull request
+            if get(ENV, "CI_PULL_REQUEST", "false") == "false"
+                data["git"]["branch"] = split(ENV["GIT_BRANCH"], "/")[2]
+            end
 
             if verbose
                 println("Submitting data to Coveralls...")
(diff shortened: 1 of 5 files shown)


src/coveralls.jl
```diff
@@ -11,7 +11,6 @@ module Coveralls
     using HTTP
     using JSON
     using Compat
-    using Compat.LibGit2
```
ararslan: Why was this removed? I'm fairly certain this is still needed.

laurentheirendt: Is `LibGit2` not part of `Base`? (https://docs.julialang.org/en/stable/devdocs/libgit2/index.html)

ararslan: It is in Julia 0.6 but not in 0.7. In 0.6, `using Compat.LibGit2` is the same as `using Base.LibGit2`, and in 0.7 it's the same as `using LibGit2`, which is part of the standard library rather than Base. Either way, this is still required.

laurentheirendt: I see ... Thanks for clarifying! I will add it back in ...

laurentheirendt:
@ararslan, may you merge this PR? Thanks 👍

PR #167, opened 2018-07-16
Fix Meta.parse call on 32-bit

omus:
The `position` function will always return an `Int64` while the `Meta.parse` only accepts an `Int`. This means that when running 32-bit Julia `Meta.parse` will only accept an `Int32` which causes this call to fail.


omus:
AppVeyor will need to be enabled for this repo.

PR #174, opened 2018-09-13
Change REPO_TOKEN to COVERALLS_TOKEN

### LINKED ISSUE #173: Change REPO_TOKEN to COVERALLS_REPO_TOKEN
Since both coveralls and codecov can co-exist simultaneously. Also, this is the name suggested in https://github.com/coveralls-net/coveralls.net/wiki/CI-Integrations#appveyor

simonbyrne:
Makes it consistent with codecov, fixes #173.


simonbyrne:
Those Appveyor test results don't look good...

fingolfin:
Could you perhaps rebase this? (Probably best to first drop the changes to  .travis.yml and appveyor.yml which are already master.

src/coveralls.jl
```diff
@@ -158,7 +166,11 @@ module Coveralls
     git_info can be either a `Dict` or a function that returns a `Dict`.
     """
     function submit_token(fcs::Vector{FileCoverage}, git_info=query_git_info; kwargs...)
-        data = Dict("repo_token" => ENV["REPO_TOKEN"],
+        data = Dict("repo_token" => get(ENV,"COVERALLS_TOKEN") do
+                            get(ENV, "REPO_TOKEN") do #backward compatibility
+                                error("Coveralls submission from AppVeyor requires a COVERALLS_TOKEN environment variable")
```
fingolfin: The "AppVeyor" here should be removed. 

Also (and I realize this is not your fault, it's in the existing code), I wonder why repo token is not supported for Travis. According to the Travis documentation, it may also be needed on Travis, namely for private repos.

The reason I bring this up is that this might be an opportunity to refactor this code.

PR #184, opened 2018-10-25
Rename Coveralls.submit_token to submit_local

fingolfin:
This matches the Codecov module


src/coveralls.jl
```diff
@@ -185,4 +185,7 @@ module Coveralls
             println(String(r.body))
         end
     end
+
+    @deprecate submit_token submit_local
```
ararslan: This will need to be `Base.@deprecate`, assuming there isn't already a `using Base: @deprecate` in this module.

fingolfin: OK... So why isn't that needed in src/codecovio.jl (where I took this from)?

fingolfin:
Travis completed, but still is shown as pending here; seems to be a glitch? Perhaps the Travis job needs to be restarted.

src/coveralls.jl
ararslan: Whoa, weird... it looks like at some point `@deprecate` got exported from Base. Not sure when that happened. Never mind then.

src/coveralls.jl
```diff
@@ -156,14 +156,15 @@ module Coveralls
...
-        submit_token(fcs::Vector{FileCoverage}, git_info=query_git_info; kwargs...)
+        submit_local(fcs::Vector{FileCoverage}, git_info=query_git_info; kwargs...)
 
     Take a `Vector` of file coverage results (produced by `process_folder`),
     and submits them to Coveralls. For submissions not from TravisCI.
 
     git_info can be either a `Dict` or a function that returns a `Dict`.
     """
-    function submit_token(fcs::Vector{FileCoverage}, git_info=query_git_info; kwargs...)
+
```
ararslan: This line break should be removed, otherwise the docstring doesn't apply to the function definition below it.

simonbyrne: I think that was due to my rebase.

PR #185, opened 2018-10-25
Rename appveyor.yml to .appveyor.yml

fingolfin:


ararslan:
It's convention to keep it as appveyor.yml with no dot, I think because once upon a time AppVeyor didn't support it with a dot (could be wrong on that). I don't know of any other repositories that use it with a dot.

fingolfin:
Whose convention is this? "It used not to work" doesn't seem relevant (it's been supported at the very least since January 2017, but I think actually quite a while longer). On the pro side, I would say: "It is consistent with putting all those other files managing CI and coverage as dot files".

Also, note that the `README.md` of this very repository suggests to use `.appveyor.yml` :-).

Other repositories using it:
* https://github.com/JuliaLang/PackageCompiler.jl
* https://github.com/gap-system/gap/

ararslan:
> Whose convention is this?

None formally, it's just what's in Example.jl, the canonical Julia example package, as well as the vast majority of existing packages.

simonbyrne:
I didn't know it worked with a dot! Apparently [it was added in 2016](https://github.com/appveyor/ci/issues/1040#issuecomment-246847714).

I think we should definitely move to that (and perhaps update Example.jl).

PR #190, opened 2018-12-05
Coverage: reduce arbitrary differences between submit and submit_local (drop pre-1.0 Compat)

vtjnash:
and simplify the code a bit for Julia v1.0


.travis.yml
```diff
@@ -1,7 +1,5 @@
 language: julia
 julia:
-  - 0.6
```
ararslan: The package still supports 0.6

vtjnash: not anymore, I used 1.0-only features :P

but seriously, do future versions need to still support 0.6?

ararslan: No, but the REQUIRE file should be updated to reflect that

vtjnash: Ah, ok, I forgot about that one. I think I got a few more references too.

PR #191, opened 2018-12-14
Add support for reading LCOV files

vtjnash:
Julia v1.1 is slated to have a new, more flexible code-coverage format (JuliaLang/julia#30381). This is the prep-work (e.g. helper functions) for utilizing that information.


ararslan:
Looks like that change likely won't make it into 1.1.

vtjnash:
I think we'll get it in. But regardless, the only part of this that depends on it is the sample usage section of the README.

vtjnash:
@fingolfin You seem to be the current de-facto maintainer—does this look good to you?

README.md
```diff
@@ -25,13 +18,8 @@ Coverage.jl
 
 *Step 1:* Navigate to your test directory, and start julia like this:
 ```sh
-julia --code-coverage=user
-```
-or, if you're running Julia 0.4 or higher,
-```sh
-julia --code-coverage=user --inline=no
+julia --code-coverage=tracefile-%p.info --code-coverage=user
```
fingolfin: I don't understand this line. Quoting `julia --help` (versoin 1.1.0-DEV from `relase-1.1`):
```
 --code-coverage={none|user|all}, --code-coverage
                           Count executions of source lines (omitting setting is equivalent to "user")
```
So is the option to specify a trace file something new, which has not yet landed in `release-1.1`? If so, shouldn't this example be qualified, and explain a bit more? (Also, do I really *have* to specify a tracefile? And does this mean I can't name a tracefile "user"? This all seems a bit odd to me).

README.md
```diff
@@ -25,13 +18,8 @@ Coverage.jl
...
 ```
-(Turning off inlining gives substantially more accurate results, but may slow down your tests.)
```
fingolfin: Is this not true anymore?

Manifest.toml
```diff
@@ -0,0 +1,98 @@
+# This file is machine-generated - editing it directly is not advised
```
fingolfin: Why is `Manifest.toml` added in the second commit, not the first?

src/Coverage.jl
```diff
@@ -5,27 +5,20 @@
 # https://github.com/JuliaCI/Coverage.jl
 #######################################################################
 module Coverage
-
-    # The code coverage results produced by Julia itself report some
-    # lines as "null" (cannot be run), when they could have been run
-    # but were not (should be 0). We use JuliaParser to augment the
-    # coverage results by identifying this code.
-    # this file coverage will be checked by test/data/expected.info
```
fingolfin: At first I was confused why this comment was removed, but apparently it only was moved down, into `amend_coverage_from_src!`, which makes sense.

test/data/Coverage.jl
```diff
@@ -1,81 +1,5 @@
 #######################################################################
-# Coverage.jl
-# Take Julia test coverage results and bundle them up in JSONs
-# https://github.com/IainNZ/Coverage.jl
+# This file is not executed, but is used as comparison input
```
fingolfin: Huh, most of `test/data/Coverage.jl` is removed in the first commit -- why? Also, `test/data/Coverage.jl.cov` is not updated to match. This seems wrong to me, but perhaps I am missing something?

README.md
vtjnash: Hasn't been true for years

test/data/Coverage.jl
vtjnash: They don't need to match. I just removed most of the content because it's distracting (confuses search/grep/me to have a large quantity of almost-code)

README.md
vtjnash: Yes, the help for it will explain more when that gets added: https://github.com/JuliaLang/julia/blob/11ce9f3fbffe55e73bfec968d970b26f8e8e1e19/src/jloptions.c#L133-L134

> shouldn't this example be qualified

I suppose, I could. Given the age of some of the other content here, I figured it's better just to be concise.

Manifest.toml
vtjnash: I messed up the git history, should be fixed now. Thanks.

REQUIRE
```diff
@@ -1,4 +0,0 @@
-julia 1.0
```
ararslan: REQUIRE files are still needed for registered packages

vtjnash: OK, I put it back

test/data/Coverage.jl
fingolfin: So I guess you point here is that `r = process_file(joinpath(datadir,"Coverage.jl"))` only really cares about `Coverage.jl.cov` and not about `Coverage.jl` itself (other than that `Coverage.jl` needs to exist?

But then the comment should say that. The current says that it is used as "comparison input", but if that was true, then surely removing most of the content of the file would break that comparison. So, why keep any content, other than a comment? E.g.
```
# This file is not used; it only exists as a placeholder so that `process_file` can
# be called on it in order to read `Coverage.jl.cov`.
```
or something like that.

test/data/Coverage.jl.cov
```diff
@@ -1,7 +1,7 @@
         - #######################################################################
-        - # Coverage.jl
-        - # Take Julia test coverage results and bundle them up in JSONs
-        - # https://github.com/IainNZ/Coverage.jl
+        - # This file is not executed, but is used as comparison input
```
fingolfin: ? Of course this file is not executed...

README.md
fingolfin: Well, but now any Julia 1.0 user trying to follow this example would get an error. Indeed, I just tried:
```
$ julia --code-coverage=tracefile-%p.info --code-coverage=user
ERROR: julia: invalid argument to --code-coverage (tracefile-%p.info)
```
As long as Julia 1.1 is not released (and I'd argue: even for quite a while after that), shouldn't the README also cate for Julia 1.0 users? So why not just
```
julia --code-coverage=user
```
as before? Is it even necessary to mention that `tracefile` at all by default? (Since I have no idea what the `tracefile` is for, and since this PR nowhere says, I can't tell).

Since you asked for my opinion: I would not merge it in this form if it was my project / my decision. As it is, though, of course you can do whatever you want.

README.md
fingolfin: Huh, weird, I thought it was relevant until a few weeks ago, see https://github.com/JuliaLang/julia/issues/28192 . Anyway, glad to hear things are better now.

src/lcov.jl
```diff
@@ -73,4 +76,68 @@ function writeline(io::IO, line::Int, count::Nothing)
...
+    source_files = FileCoverage[]
+    files = readdir(folder)
+    for file in files
+        fullfile = joinpath(folder, file)
+        if isfile(fullfile)
+            # Is it a tracefile?
+            if endswith(fullfile, ".info")
+                append!(source_files, readfile(fullfile))
+            else
+                println("Coverage.process_folder: Skipping $file, not a .info file")
```
fingolfin:
```suggestion
                println("Coverage.readfolder: Skipping $file, not a .info file")
```

src/lcov.jl
```diff
@@ -73,4 +76,68 @@ function writeline(io::IO, line::Int, count::Nothing)
...
+                end
+            end
+        end
+        nothing
+    end
+    return source_files
+end
+
+"""
+    readfolder(folder) -> Vector{FileCoverage}
```
fingolfin: Nitpick, but: should it be `read_file` and `read_folder` to match `process_file` and `process_folder`?

test/runtests.jl
```diff
@@ -35,44 +40,57 @@ end
...
         # we only have a single file, but we want to test on the Vector of file results
         LCOV.write(lcov, FileCoverage[r])
-        expected = read(joinpath(datadir, "expected.info"), String)
+        expected = read(joinpath(datadir, "tracefiles", "expected.info"), String)
         if Sys.iswindows()
             expected = replace(expected, "SF:test/data/Coverage.jl\n" => "SF:test\\data\\Coverage.jl\n")
         end
         @test String(take!(lcov)) == expected
+        # test that reading the LCOV file gives the same data
+        lcov = LCOV.readfolder(datadir)
```
fingolfin: So... how is one supposed to use `readfolder` later on? Call it in addition (or instead of) `process_folder` from the CI scripts (but then, why isn't `etc/travis-coverage.jl` modified to do so)? Or is this code simply not yet done, and hence not hooked up in `process_folder`? Or what?

README.md
vtjnash: It was fixed by #18196, the latest issue were just a regression from introduced the new SSAIR format and is re-fixed by #30251

src/lcov.jl
vtjnash: I was matching the existing `LCOV.writefile` with the name `LCOV.readfile`, and the general base guidelines to avoid underscores. I agree it would be nice to be more consistent, though I'm not sure how to accomplish that.

test/runtests.jl
vtjnash: In the readme, I show using both now (in tandem), with a few comments mentioning the applicability of each line. The CI script doesn't need to change, since, for most users, this addition is 6 of one or half a dozen of the other. It really only affects people who are using it locally (don't have a CI sandbox environment) or are measuring Base and stdlib coverage.

vtjnash:
@fingolfin updated per review comments, is this good now?

src/lcov.jl
fingolfin: OK, I see -- yeah, when two differing "conventions" clash, this is always a bit nasty. In any case, it doesn't matter much, I was just wondering.

test/runtests.jl
fingolfin: OK

fingolfin:
Thank you, this looks good to me now!

PR #193, opened 2019-01-15
Fixed test that failed on Windows due to line endings

zundertj:
The Windows line endings show up as `\r\n` on my machine in `expected`, while the `IOBuffer` returns `\n`. Not sure why this does not fail for others/AppVeyor, though.


vtjnash:
We typically try to demand that git checkouts are binary-preserving, as really the only sane option (https://github.com/JuliaLang/julia/blob/master/README.windows.md#line-endings)

It appears that appveyor does the same: https://www.appveyor.com/docs/windows-images-software/#version-control-systems

zundertj:
Got it, thank you for the links. Will close.

Issue #92, opened 2015-09-22
Output is pretty verbose


ssfrr:
Currently there's quite a bit of output to STDOUT even when everything is successful. Would you accept a PR adding a `verbose` flag to `process_folder` and `clean_folder` that defaults to `false`? The behavior would be that if `verbose=false`, it only prints output when something unexpected happens.

PR #156, opened 2018-01-25
Use logging infrastructure instead of prints

ararslan:

PR #194, opened 2019-01-15
Use logging framework for STDOUT calls

zundertj:
I hope you are oke with basically redoing https://github.com/JuliaCI/Coverage.jl/pull/156? Not sure Manifest changes should be in here, they were added when adding Logging.

Effectively also closes https://github.com/JuliaCI/Coverage.jl/issues/92.

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -125,8 +125,8 @@ module Coverage
         if isempty(files)
             # ... we will assume that, as there is a .jl file, it was
             # just never run. We'll report the coverage as all null.
-            println( """Coverage.process_cov: Coverage file(s) for $filename do not exist.
-                                              Assuming file has no coverage.""")
+            @info """Coverage.process_cov: Coverage file(s) for $filename do not exist.
+                                              Assuming file has no coverage."""
             nlines = countlines(filename)
             return fill!(Vector{CovCount}(undef, nlines), nothing)
         end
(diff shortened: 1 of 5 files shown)


zundertj:
I have resolved the merge conflicts. I do see there is an error related to MbedTLS, which I believe has nothing to do with this PR. Could anyone review this PR, please?

Project.toml
```diff
@@ -7,6 +7,7 @@ version = "0.8.1"
 HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
+Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
```
ararslan: The dependency on Logging (and by extension the `using Logging` occurrences) isn't actually necessary, as the basic logging macros—`@info`, `@warn`, and `@error`—are available in Base.

ararslan:
> I do see there is an error related to MbedTLS, which I believe has nothing to do with this PR

Correct, that's https://github.com/JuliaWeb/MbedTLS.jl/issues/193.

Project.toml
zundertj: Got it, removed the entry and the using statements.

PR #195, opened 2019-01-16
Factored out create_uri_string_* methods for easier testing

zundertj:
Removes the need for Suppressor, and better shows what is actually tested with the dry run, namely generating the token. When changing this code, I also realized that `submit_generic()`, although exported, is probably not used outside the package? I.e., for CI one would use `submit()`, and for local `submit_local()`. Should that export be removed?

### DIFF
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -65,6 +65,11 @@ module Codecov
     on TravisCI or AppVeyor. If running locally, use `submit_local`.
     """
     function submit(fcs::Vector{FileCoverage}; kwargs...)
+        submit_generic(fcs; add_ci_to_kwargs(;kwargs...)...)
+    end
+
+
+    function add_ci_to_kwargs(;kwargs...)
         if lowercase(get(ENV, "APPVEYOR", "false")) == "true"
             appveyor_pr = get(ENV, "APPVEYOR_PULL_REQUEST_NUMBER", "")
             appveyor_job = join(
@@ -125,9 +130,8 @@ module Codecov
             error("No compatible CI platform detected")
         end
 
-        submit_generic(fcs; kwargs...)
+        return kwargs
     end
-
 
     """
         submit_local(fcs::Vector{FileCoverage}, dir::AbstractString=pwd())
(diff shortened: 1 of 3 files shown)


src/codecovio.jl
```diff
@@ -65,6 +65,94 @@ module Codecov
...
+        submit_generic(fcs::Vector{FileCoverage})
+
+    Takes a vector of file coverage results (produced by `process_folder`),
+    and submits them to a Codecov.io instance. Keyword arguments are converted
+    into a generic Codecov.io API uri.  It is essential that the keywords and
+    values match the Codecov upload/v2 API specification.
+    The `codecov_url` keyword argument or the CODECOV_URL environment variable
+    can be used to specify the base path of the uri.
+    """
+    function submit_generic(fcs::Vector{FileCoverage}; kwargs...)
```
fingolfin: Is there a reason why you are reordering the code? It makes it much harder to review this PR!

zundertj: You are right, I changed the order of some functions. I have been able to reduce that a bit by moving `construct_uri_string()` to the bottom of the file. Still, as I'm basically disentangling two functions, there is inevitably a number of lines being moved.

fingolfin: Thanks, that's much better. Of course I realize you can't fully avoid moving some stuff around, but reducing it to a minimum (resp. performing the moving in a separate commit) helps a lot in reviewing. I took the liberty to move some more stuff down, and rebase it to resolve conflicts, squashing it into one commit. Let's see if the tests pass.

src/codecovio.jl
```diff
@@ -65,6 +65,12 @@ module Codecov
     on TravisCI or AppVeyor. If running locally, use `submit_local`.
     """
     function submit(fcs::Vector{FileCoverage}; kwargs...)
+        submit_generic(fcs; add_ci_to_kwargs(;kwargs...)...)
+    end
+
+
+    function add_ci_to_kwargs(;kwargs...)
+        @show kwargs
```
fingolfin: That looks like it should be removed, no?

src/codecovio.jl
```diff
@@ -169,12 +178,34 @@ module Codecov
     values match the Codecov upload/v2 API specification.
     The `codecov_url` keyword argument or the CODECOV_URL environment variable
     can be used to specify the base path of the uri.
-    The `dry_run` keyword can be used to prevent the http request from
-    being generated
```
fingolfin: Why is this removed?

zundertj:
I have fixed the remarks you made. I think initially I wanted to remove the `dry_run` keyword argument altogether, as it's main use seems to be to verify that the uri string (now separated out in the function `create_uri_string`) was created correctly, there is not much else going on in `submit_generic` outside the `if`-statement on `dry_run` (main purpose the tests?). Obviously that would break the API, so it is not great to remove it.

Issue #180, opened 2018-10-24
Setup tests for Coveralls integration, fix submit_token and query_git_info


fingolfin:
For Codecov integration, there are several offline tests in `test/runtests.jl`, but not so for Coveralls. This should be changed. Esp. since I suspect some of the currently untested functions, like `submit_token`, are currently broken (`submit_token` should likely receive the same fix as `submit` for it use of `HTTP.post`; see issue #175). I'd also rename `submit_token` to `submit_local` as that (a) matches Codecov and (b) is self-explanatory, unlike `submit_token` (which does not submit a token, after all.

Moreover, `Coveralls.query_git_info` seems to be broken in Julia >= 0.6:
```
julia> using Coverage; Coveralls.query_git_info()
ERROR: GitError(Code:ENOTFOUND, Class:Config, Remote 'mh/fix' does not exist.)
Stacktrace:
 [1] macro expansion at ./libgit2/error.jl:99 [inlined]
 [2] get(::Type{Base.LibGit2.GitRemote}, ::Base.LibGit2.GitRepo, ::String) at ./libgit2/remote.jl:68
 [3] query_git_info(::String) at /Users/mhorn/.julia/v0.6/Coverage/src/coveralls.jl:127
 [4] query_git_info() at /Users/mhorn/.julia/v0.6/Coverage/src/coveralls.jl:113
```
Indeed, the call `LibGit2.get(LibGit2.GitRemote, repo, branch)` seems to be wrong (but `LibGit2.get` is undocumented, so I am not 100% sure).

fingolfin:
Most of this was done in PR #196, "only" the tests for Coveralls integration are missing; to reduce confusion, I put that into a new issue #206.

PR #196, opened 2019-01-16
Coveralls query_git_info() fix + test + small refactoring

zundertj:
Covers most of https://github.com/JuliaCI/Coverage.jl/issues/180, except for the end-to-end test of `submit()` and `submit_local()`.

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -78,28 +78,9 @@ module Coveralls
             error("No compatible CI platform detected")
         end
 
-        repo_token =
-                get(ENV, "COVERALLS_TOKEN") do
-                    get(ENV, "REPO_TOKEN") do #backward compatibility
-                        if data["service_name"] != "travis-ci"
-                            error("Coveralls submission requires a COVERALLS_TOKEN environment variable")
-                        end
-                    end
-                end
-        if repo_token != nothing
-            data["repo_token"] = repo_token
-        end
+        data = add_repo_token(data, false)
 
-        if verbose
-            @info "Submitting data to Coveralls..."
-        end
-
-        req = HTTP.post("https://coveralls.io/api/v1/jobs", HTTP.Form(makebody(data)))
-
-        if verbose
-            @info "Result of submission:\n" * String(req.body)
-        end
-        return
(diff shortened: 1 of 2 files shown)


src/coveralls.jl
```diff
@@ -78,18 +78,11 @@ module Coveralls
...
-                            error("Coveralls submission requires a COVERALLS_TOKEN environment variable")
-                        end
-                    end
-                end
-        if repo_token != nothing
-            data["repo_token"] = repo_token
-        end
+        data = add_repo_token(data, false)
 
+<<<<<<< HEAD
```
ararslan: You have some merge conflict markers committed here.

fingolfin: Basically, you may want to drop that last commit, and instead rebase your PR onto the latest master.

While at it, you have a commit "Fixed faulty previous commit, factored out add_repo_token" -- perhaps just squash that into the preceding commit (using an interactive rebase, `git rebase -i`, quite handy tool); similarly, the commit "Fixed bug in call to add_repo_token" could be squashed into the second commit.

If you don't know how to do any of these, but are interested in learning, I'd be happy to coach you through it, too. It really isn't hard.

zundertj: I wasn't sure whether you'd want me to squash, it can be confusing when reviewing. I have now squashed all commits together.

src/coveralls.jl
```diff
@@ -173,14 +152,41 @@ module Coveralls
...
-        return
     end
 
+    # adds the repo token to the data
+    function add_repo_token(data, local_submission)
+        repo_token =
+                get(ENV, "COVERALLS_TOKEN") do
+                    get(ENV, "REPO_TOKEN") do #backward compatibility
+                        # error unless we are on Travis
+                        if local_submission | (data["service_name"] != "travis-ci")
```
fingolfin: Why `|` ? Shouldn't this be `||`

zundertj: You are right, I fixed this.

zundertj:
Thank you for rebasing. I have fixed the `|`, and this should now be good to go.

PR #197, opened 2019-01-24
Fix Coveralls submission with HTTP 0.8

ararslan:


ararslan:
Hm, some kind of version resolution issue; we need HTTP 0.8 but something is keeping us from installing it.

src/coveralls.jl
```diff
@@ -94,10 +94,7 @@ module Coveralls
             println("Submitting data to Coveralls...")
         end
 
-        url = "https://coveralls.io/api/v1/jobs"
-        body = HTTP.Form(makebody(data))
-        headers = ["Content-Type" => "multipart/form-data; boundary=$(body.boundary)"]
-        req = HTTP.post(url, headers, body)
+        req = HTTP.post("https://coveralls.io/api/v1/jobs", HTTP.Form(makebody(data)))
```
vtjnash: missing an identical update later in the file

ararslan: Good catch, thanks!

REQUIRE
```diff
@@ -1,4 +1,4 @@
 julia 1.0
```
HarrisonGrodin: Afaict, Coveralls still fails on 0.7, I assume due to a versioning issue. Should this be changed to `0.7`?

PR #201, opened 2019-02-18
Only add lines from function bodies for coverage

### LINKED ISSUE #199: Function signature is reported as uncovered
I have lots of functions that look like this:

```julia
function Base.show(io::IO, ::MIME"text/html", source::StatFile)
    TableShowUtils.printHTMLtable(io, getiterator(source))
end
```

The ``*.cov`` file part for this one looks as follows, both on julia 1.0 and 1.1:

```julia
        - function Base.show(io::IO, ::MIME"text/html", source::StatFile)
        6     TableShowUtils.printHTMLtable(io, getiterator(source))
        - end
```

When I process this with Coverage.jl, then the first line here gets marked as code that is not covered, which seems wrong.

I think this is distinct from the inline issue discussed in https://github.com/JuliaCI/Coverage.jl/issues/187.

I'm wondering now where this goes wrong... Should Base report this first line as covered? There is actually code there that is running, namely the ``MIME"text/htm"`` part, right?

davidanthoff:
Fixes #199.

There are certain function signatures that have a line node in their AST, and then the function signature gets marked as uncovered. That is not good, because there is then no way to actually cover that line. So this just skips the function signature part, and only looks at the body of functions when it comes to amend the line info reported from base.

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -3,6 +3,9 @@
 isevaldef(x) = Base.Meta.isexpr(x, :(=)) && Base.Meta.isexpr(x.args[1], :call) &&
                x.args[1].args[1] == :eval
 
+# This detects two types of function declarations: those that use the
+# `function` keyword (first case) and those that are defined as
+# `f() = expr` (the second case).
 isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
@@ -33,9 +36,21 @@ function function_body_lines!(flines, ast::Expr, infunction)
         args = ast.args
     end
 
-    if isfuncexpr(ast) && length(args)>=2 && args[2] isa Expr
-        # Only look in function body
-        for arg in args[2].args
+    # Check whether we are looking at a function declaration.
+    # Then also make sure we are not looking at a function declaration
+    # that does not also define a method by checking the length of
+    # ast.args and making sure it is at least length==2. The test for
+    # Expr might not be necessary but also can't harm. Note that this works
+    # for both function declarations that use the `function` keyword, and
+    # declarations of the form `foo() = expr`, in both cases the method
+    # body ends up being `ast.args[2]` (if it exists).
+    if isfuncexpr(ast) && length(ast.args)>=2 && ast.args[2] isa Expr
+        # Only look in function body and ignore the function signature
+        # itself. Sometimes function signatures have line nodes inside
+        # and we don't want those lines to be identified as runnable code.
+        # In this context, ast.args[1] is the function signature and
+        # ast.args[2] is the method body
+        for arg in ast.args[2].args
             flines = function_body_lines!(flines, arg, true)
         end
     else
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -32,9 +32,17 @@ function function_body_lines!(flines, ast::Expr, infunction)
     else
         args = ast.args
     end
-    infunction |= isfuncexpr(ast)
-    for arg in args
-        flines = function_body_lines!(flines, arg, infunction)
+
+    if isfuncexpr(ast) && length(args)>=2 && args[2] isa Expr
+        # Only look in function body
+        for arg in args[2].args
+            flines = function_body_lines!(flines, arg, true)
+        end
+    else
+        for arg in args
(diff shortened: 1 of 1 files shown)


davidanthoff:
@fingolfin Could you review this? I believe you added the original logic here, so it would be good to make sure this PR doesn't overlook something important. Thanks!

davidanthoff:
Bump, would be great if this could be reviewed & merged soon!

src/parser.jl
```diff
@@ -32,9 +32,17 @@ function function_body_lines!(flines, ast::Expr, infunction)
     else
         args = ast.args
     end
-    infunction |= isfuncexpr(ast)
-    for arg in args
-        flines = function_body_lines!(flines, arg, infunction)
+
+    if isfuncexpr(ast) && length(args)>=2 && args[2] isa Expr
```
fingolfin: Could you perhaps add a comment that explains what's going on? (No worries, I won't hold this PR hostage if you don't have time to do that, given that the rest of the code is not exactly a shining beacon for well-commented code to start with... but since you clearly understand what you are doing, it might not be too hard for you to write something, and it'd help others who are trying to figure out what this code does in the future...

E.g. people like me... I'll now go to read up on this code to figure out what the "length(args)>=2 && args[2] isa Expr" bit is about and then come back to conduct a proper review ;-).

fingolfin: OK, looking at https://docs.julialang.org/en/v1/devdocs/ast/ it says:
> A basic function definition is parsed as (function (call f x) body). 

So I guess the code above is meant to check if there is a function body, and in that case, skip the arguments?
But is there actually a case when length(args) is different from 2?  Hm, perhaps in that weird second type of ast expressions for which `isfuncexpr` returns true? Can somebody explain that bit of `isfuncexpr` to me? I.e. the bit after `ex.head == :function || ` in the following definition:
```julia
isfuncexpr(ex::Expr) =
    ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
```

davidanthoff: @fingolfin I've added a whole bunch of comments. If something is still unclear, let me know!

src/parser.jl
```diff
@@ -3,6 +3,9 @@
 isevaldef(x) = Base.Meta.isexpr(x, :(=)) && Base.Meta.isexpr(x.args[1], :call) &&
                x.args[1].args[1] == :eval
 
+# This detects two types of function declarations: those that use the
+# `function` keyword (first case) and those that are defined as
+# `f() = expr` (the second case).
```
fingolfin: Ahhh, of course. Thanks, that's helpful!

src/parser.jl
```diff
@@ -32,9 +35,29 @@ function function_body_lines!(flines, ast::Expr, infunction)
     else
         args = ast.args
     end
-    infunction |= isfuncexpr(ast)
-    for arg in args
-        flines = function_body_lines!(flines, arg, infunction)
+
+    # Check whether we are looking at a function declaration
```
fingolfin:
```suggestion
    # Check whether we are looking at a function declaration.
```

davidanthoff:
Alright, ready to merge? And tag a release :)

fingolfin:
Nope, tests are failing now (due to a typo?)

src/parser.jl
```diff
@@ -32,9 +35,29 @@ function function_body_lines!(flines, ast::Expr, infunction)
...
+    # declarations of the form `foo() = expr`, in both cases the method
+    # body ends up being `ast.args[2]` (if it exists).
+    if isfuncexpr(ast) && length(ast.args)>=2 && ast.args[2] isa Expr
+        # Only look in function body and ignore the function signature
+        # itself. Sometimes function signatures have line nodes inside
+        # and we don't want those lines to be identified as runnable code.
+        # In this context, ast.args[1] is the function signature and
+        # ast.args[2] is the method body
+        for arg in ast.args[2].args
+            flines = function_body_lines!(flines, ast.arg, true)
```
fingolfin:
```suggestion
            flines = function_body_lines!(flines, arg, true)
```

davidanthoff: Ah, thanks, that was a stupid regression :)

fingolfin:
Thanks!

davidanthoff:
@fingolfin Thanks! Could you also tag a release?

PR #203, opened 2019-03-03
Update README.md

fingolfin:


README.md
```diff
@@ -88,113 +88,155 @@ When using Coverage.jl locally, over time a lot of `.cov` files can accumulate.
...
+     line downloads this package, collects the per-file coverage data, then
+     bundles it up and submits to Coveralls. Coverage.jl assumes that the
+     working directory is the package directory, so it changes to that first
+     (so don't forget to replace `MyPkg` with your package's name!
+
+   - On Travis CI:
+
+       ```yml
+       after_success:
+       - julia -e 'cd(Pkg.dir("MyPkg")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
```
ararslan: `cd(Pkg.dir("..."))` is for Julia 0.6 and prior; for 0.7 and later, that step should be replaced with `using Pkg`.

fingolfin: This code is indeed missing a `using Pkg`, thanks for catching it. The `cd` is indeed not needed for packages using a `Project.toml`. For older packages, I think it might be needed -- but since of course this README is only interesting for new packages, and those should all be using `Project.toml`, So overall, I agree, and pushed a corresponding change. Thank you!

Issue #175, opened 2018-09-26
Coveralls integration stopped working around 2018-09-20


maxbennedich:
This seems to be a global issue, I've checked several packages, and they all started failing around Sept 20. This is what happens during submission of coverage:
```
Submitting data to Coveralls...
┌ Warning: Error in @async writebody task
│   exception =
│    MethodError: no method matching write(::HTTP.Streams.Stream{HTTP.Messages.Response,HTTP.ConnectionPool.Transaction{MbedTLS.SSLContext}}, ::Pair{String,HTTP.Multipart{Base.GenericIOBuffer{Array{UInt8,1}}}})
...
 [21] #submit#1(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Array{FileCoverage,1}) at /home/travis/.julia/packages/Coverage/25tif/src/coveralls.jl:91
 [22] submit(::Array{FileCoverage,1}) at /home/travis/.julia/packages/Coverage/25tif/src/coveralls.jl:60
```
See example of full stack trace in [this build](https://travis-ci.org/JuliaData/DataFrames.jl/jobs/433126893#L1380).

Codecov still works.

vchuravy:
x-ref: https://github.com/JuliaWeb/HTTP.jl/issues/245#issuecomment-414847061

samoconnor:
Passing a `HTTP.Form` as the body and setting the `Content-Type` should do the trick:
```julia
using HTTP
using JSON

makebody(data::Dict) =
    HTTP.Form(
    Dict("json_file" => HTTP.Multipart("json_file", IOBuffer(JSON.json(data)),
                                       "application/json")))

url = "https://coveralls.io/api/v1/jobs"

data = Dict("service_job_id"    => "APPVEYOR_JOB_ID",
            "service_name"      => "appveyor",
            "source_files"      => "xyz",
            "repo_token"        => "REPO_TOKEN")

body = makebody(data)
headers = ["Content-Type" => "multipart/form-data; boundary=$(body.boundary)"]

req = HTTP.post(url, headers, body, verbose=3)
```
```
DEBUG: 2018-10-02T08:04:05.335 6516 ➡️  "POST /api/v1/jobs HTTP/1.1\r\n" (write)
[...]

PR #204, opened 2019-03-04
Restore HTTP.post invocation

fingolfin:
In PR #196, part of the changes in PR #197 were reverted, I think by accident (or rather, due to my fault, when rebasing the former I was unaware of the changes in PR #197 which had already been merged some time ago).

This PR restores this. However, I am not 100% sure this is even right; there was a reason we added those http headers. I am assuming the point behind PR #197 was that this is not necessary anymore -- but neither the description of that PR, nor the commits in it, actually explain this (only says this is meant to "fix compatibility with HTTP 0.8 -- but curiously, merging PR #196 did not seem to break any tests?).

So before merging this, I'd like to verify that the cause for the addition of those explicit headers is indeed gone now (discussion of that is on PR #175). I am assuming it is the case, though, due to https://github.com/JuliaWeb/HTTP.jl/commit/01a2cff816e3cd6095990b80167f216cb26e906c ...

Anyway, @ararslan will know and then can decide whether to merge this PR here or not :-).

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -161,10 +161,7 @@ module Coveralls
             @info "Submitting data to Coveralls..."
         end
 
-        url = "https://coveralls.io/api/v1/jobs"
-        body = HTTP.Form(makebody(data))
-        headers = ["Content-Type" => "multipart/form-data; boundary=$(body.boundary)"]
-        req = HTTP.post(url, headers, body)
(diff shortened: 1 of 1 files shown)


ararslan:
Yes, `HTTP.Form` now adds the headers that we were previously adding manually as of HTTP 0.8, as you noted in the linked HTTP commit. I think the changes from #197 are still relevant.

PR #208, opened 2019-03-06
Check DISABLE_AMEND_COVERAGE_FROM_SRC env var

fingolfin:
If the environment variable DISABLE_AMEND_COVERAGE_FROM_SRC is set to `yes`, do not call `amend_coverage_from_src!`. This is an optional mitigation for issue #187, and in general makes it easier to test the impact of amending the coverage data.

PR #210, opened 2019-03-07
Improve coverage heuristic in function_body_lines

fingolfin:
Improve coverage heuristic in function_body_lines

The purpose of function_body_lines is to find functions which are "dead" (= never executed by test suite), and mark them as code, because Julia's built-in coverage code does not do that; if we were to rely on it, such functions would be marked as non-code, and thus they do not show up in the overall code coverage percentage.

But our current heuristic is too aggressive in so far as that Julia also sometimes marks code lines as "is not code" in the middle of functions that are being executed (most likely due to inlining and other optimizations). If we mark those lines as code, they will show up as uncovered, with no way for the package author to change this.

With this commit, we refine our heuristic: we only apply it to functions for which Julia recorded no coverage at all, as before; but if even just one line in the function body has coverage information, we do not touch the code lines in it at all.

Some care must be taken to make sure this works right for nested functions.

Closes #188

### DIFF
--- a/src/Coverage.jl
+++ b/src/Coverage.jl
@@ -175,7 +175,7 @@ module Coverage
             linestart = minimum(searchsorted(linepos, pos - 1))
             ast, pos = Meta.parse(content, pos)
             isa(ast, Expr) || continue
-            flines = function_body_lines(ast)
+            flines = function_body_lines(ast, coverage, linestart - 1)
             if !isempty(flines)
                 flines .+= linestart-1
                 for l in flines
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -10,16 +10,19 @@ isfuncexpr(ex::Expr) =
     ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
 isfuncexpr(arg) = false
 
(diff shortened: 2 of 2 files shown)


fingolfin:
Nice, this seems to work, and thus improves coverage even for Coverage.jl; but it's not yet "perfect", e.g. https://codecov.io/gh/JuliaCI/Coverage.jl/src/f06e6fe198fa14e71e92eabcba51da77f5c25644/src/lcov.jl#L72 still gets marked as code; I am not quite sure why.

So even with this, there is probably still some need for PR #208, but hopefully less so.

ararslan:
I don't understand this part of the code well enough to provide a particularly meaningful review, but for what it's worth, nothing looks blaringly wrong to me. 😛 Perhaps @vtjnash would be able to provide more insight?

fingolfin:
Of course if something is wrong (blaringly or not), I'll work on fixing it (or at worst, we can revert the change again).

davidanthoff:
Is there any hope that at some point julia will just output the correct info here, so that Coverage.jl doesn't have to patch things up? Is there a julia issue tracking this?

fingolfin:
@davidanthoff as discussed on issue #187, the relevant Julia issue is 
https://github.com/JuliaLang/julia/issues/28192

PR #213, opened 2019-03-07
Add analyze_malloc test with actual .mem file.

### LINKED ISSUE #211: Add test inputs for `analyze_malloc_files`
Function `analyze_malloc_files` is currently mostly not tested. To fix that, we should add some `*.mem` file to `test/data`. But I have no idea how to generate those. If anybody wants to chip in, a PR would be most welcome.

zundertj:
Closes #211.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -40,8 +40,12 @@ end
         datadir = joinpath("test", "data")
         # Process a saved set of coverage data...
         r = process_file(joinpath(datadir, "Coverage.jl"))
+
         # ... and memory data
-        analyze_malloc(datadir)
+        malloc_results = analyze_malloc(datadir)
+        filename = joinpath(datadir, "testparser.jl.9172.mem")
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -40,8 +40,16 @@ end
...
         r = process_file(joinpath(datadir, "Coverage.jl"))
+
         # ... and memory data
-        analyze_malloc(datadir)
+        malloc_results = analyze_malloc(datadir)
+        if Sys.iswindows()
+            filename = "test\\data\\testparser.jl.9172.mem"
+        else
+            filename = "test/data/testparser.jl.9172.mem"
+        end
```
ararslan: You can just make this
```julia
filename = joinpath(@__DIR__, "data", "testparser.jl.9172.mem")
```
That takes care of the path separator differences across platforms.

zundertj: Yes, that is much cleaner. Updated and force-pushed.

PR #215, opened 2019-03-10
Handle more types of function declarations

fingolfin:
Motivated by parts of PR #214.

A real life example of dead code not found because of this can be seen [here in the JuliaSpace/ReferenceFrameRotations.jl project](https://codecov.io/gh/JuliaSpace/ReferenceFrameRotations.jl/src/master/src/quaternion.jl#L77) (CC @ronisbr). Just in case the line numbers change, the function in question is this
```julia
Quaternion{T}(u::UniformScaling) where T = Quaternion{T}(T(u.λ), T(0), T(0), T(0))
```

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -3,11 +3,15 @@
 isevaldef(x) = Base.Meta.isexpr(x, :(=)) && Base.Meta.isexpr(x.args[1], :call) &&
                x.args[1].args[1] == :eval
 
-# This detects two types of function declarations: those that use the
-# `function` keyword (first case) and those that are defined as
-# `f() = expr` (the second case).
+# This detects different types of function declarations:
+# - those that use the  `function` keyword;
+# - those that are defined as `f = x -> expr`
+# - those that are defined as `f() = expr`
+# - those that are defined as `f() where {T} = expr`
 isfuncexpr(ex::Expr) =
-    ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
+    ex.head == :function || ex.head == :-> ||
+        (ex.head == :(=) && typeof(ex.args[1]) == Expr &&
+            (ex.args[1].head == :call || (ex.args[1].head == :where && ex.args[1].args[1].head == :call)))
 isfuncexpr(arg) = false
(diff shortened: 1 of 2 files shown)


test/runtests.jl
```diff
@@ -31,8 +31,15 @@ end
     @test !Coverage.iscovfile("/otherdir/test.jl.cov", "/somedir/test.jl")
 end
 
-@testset "isfuncexpr" begin
```
ararslan: Did you mean to remove the test set here?

test/runtests.jl
```diff
@@ -31,8 +31,15 @@ end
...
+@test Coverage.isfuncexpr(:(f() = x))
```
ararslan: Should be indented to match the others inside of the test set (assuming you did not intend to remove that)

test/runtests.jl
fingolfin: Nope, fixed

ronisbr:
Awesome! I was about to ask if anyone has any clue why this line is being ignored :)

src/parser.jl
```diff
@@ -3,11 +3,15 @@
...
+# This detects different types of function declarations:
+# - those that use the  `function` keyword;
+# - those that are defined as `f = x -> expr`
+# - those that are defined as `f() = expr`
+# - those that are defined as `f() where {T} = expr`
 isfuncexpr(ex::Expr) =
-    ex.head == :function || (ex.head == :(=) && typeof(ex.args[1]) == Expr && ex.args[1].head == :call)
+    ex.head == :function || ex.head == :-> ||
+        (ex.head == :(=) && typeof(ex.args[1]) == Expr &&
+            (ex.args[1].head == :call || (ex.args[1].head == :where && ex.args[1].args[1].head == :call)))
```
vtjnash: Could also use `Base.is_short_function_def` for part of this (to handle a few more cases, including `f() where A where B = c` and `()::C = c`)

fingolfin: Cool, didn't know about `Base.is_short_function_def`. This indeed handles `f() where A where B = c`. However, it does not seem to recognize `()::C = c` as a function definition either.

PR #218, opened 2019-03-10
Skipping files and lines.

### LINKED ISSUE #44: Excluding some files from coverage
Hello,

Is there a way to exclude a file from coverage?

It's a file that generates a bunch of symbols using a macro, and I've tried several ways to make it fall under test coverage but none of them have worked, unfortunately.

### LINKED ISSUE #53: Have an option to mark lines as "ok that this isn't covered"
I have a situation where I'm calling an external function that returns an int. I know all the values that this external function can return, but I want to make sure my code breaks in case the external function adds another, new return type. So my code looks like this:

```
ret = somef()
if ret==1
  # do something
elseif ret==2
  # do something
else
  error("This cannot happen with the current version of somef, but might in the future")
end
```

It would be great if there as a way to mark that line in the else block as "all good that this is not covered", so that it doesn't show up as a red line in the coverage UI.

zundertj:
Work in progress; just setting this up to get your input.

Adresses #53 & #44; supersedes #55.

Main design issues/questions/todo's:
- [x] should a `.coverage.yml` file only affect files in the folder, or also sub folders? I guess most packages have just a single src folder, but having to specify the exclusion pattern for each and every folder in case there are multiple seems tedious to me, in particular for the line pattern. Alternatively, a much simpler solution would be to simply have a toggle or default regex in the API, removing the need for the `.coverage.yml` file for the line patterns altogether. That does not allow one to completely exclude files tough.
- [x] The current implementation fails as `amend_coverage_from_src` modifies the results afterwards. There is more code change coming for that function (PR214), so I did not want to touch that function right now. To make this work, it seems that we have to reorganize these two functions. I'm happy to do that once the dust is settled, if you agree that is the right approach.
- [x] add whatever implementation we end up with to the README. 

Fixes #44
Fixes #53
Closes #55


vtjnash:
> The current implementation fails as amend_coverage_from_src modifies the results afterwards

Due to this, I think it would be better for this implementation to live inside that function entirely. After all, this is exactly ‘amending coverage from looking at the source code’.

zundertj:
Resolved conflicts and moved the line-by-line functionality to `amend_coverage_from_src`. The only question left is whether we should apply config files also to sub folders, at the moment it does not.

src/Coverage.jl
```diff
@@ -201,24 +205,34 @@ module Coverage
                 end
             end
         end
+
+        # Verify that there is no line exclusion tag
+        i = 0  # counter to keep track of line number as in original file
+        for line in eachline(IOBuffer(content))
+            i += 1
+            if any(occursin.(excluded_line_patterns, Ref(line[10:end])) .!= nothing)
```
vtjnash:
```suggestion
            if any(pat -> occursin(pat, line), excluded_line_patterns)
```

src/Coverage.jl
```diff
@@ -136,28 +137,31 @@ module Coverage
...
-                # Columns 1:9 contain the coverage count
-                cov_segment = line[1:9]
                 # If coverage is NA, there will be a dash
-                push!(coverage, cov_segment[9] == '-' ? nothing : parse(Int, cov_segment))
+                if line[9] == '-'
+                    push!(coverage, nothing)
+                else
+                    # Columns 1:9 contain the coverage count
+                    push!(coverage, parse(Int, line[1:9]))
+                end
```
fingolfin: This part of the changes seems to be purely cosmetically, or am I missing something?

src/Coverage.jl
```diff
@@ -232,27 +246,80 @@ module Coverage
...
             end
         end
         return source_files
     end
 
+    """
+        get_exclusions(file::AbstractString) -> (excluded_files, excluded_line_patterns)
+
+    Parse the provided file as a coverage exclusions file (usually named
+    ``.coverage.yml`). This file should have a format like the following:
```
fingolfin:
```suggestion
    `.coverage.yml`). This file should have a format like the following:
```

src/Coverage.jl
```diff
@@ -232,27 +246,80 @@ module Coverage
     child folders. Default folder is "src", which is useful for the primary case
     where Coverage is called from the root directory of a package.
     """
-    function process_folder(folder="src")
+    function process_folder(folder="src", exclusions_file=".coverage.yml")
```
fingolfin: What is the motivation for making the name of the YAML filed configurable. What are situations where you would expect that you'd use this?

test/runtests.jl
```diff
@@ -137,6 +137,26 @@ end
         emptycov = FileCoverage("", "", [])
         @test get_summary(emptycov) == (0, 0)
 
+        #json_data = Codecov.build_json_data(Codecov.process_folder("data"))
+        #@test typeof(json_data["coverage"]["data/Coverage.jl"]) == Array{Union{Int64,Nothing},1}
```
fingolfin: what are those commented out lines about?

src/Coverage.jl
```diff
@@ -232,27 +246,80 @@ module Coverage
...
+    ``.coverage.yml`). This file should have a format like the following:
+    ```
+    # exclude files in the directory with these names
+    exclude_files:
+      - foo.jl
+      - bar.jl
+
+    # exclude any lines that match any of the following regexes
+    exclude_patterns:
+      - "[NO COVERAGE]"
```
fingolfin: Hmm, so if I want to use this, I have to keep remember to put such a file in every single directory containing source code? If the exclusion file applied recursively, wouldn't that be more convenient (and close to how e.g. `.gitignore` works) ?

Or, put another way: Personally, I'd be content with a single master config file in the repository root; I can see imagine that some people might want to be able to override it for some subdirs (though I can't imagine *why*, but that just shows my lack of imagination :-). But having to put this file in every subdir? I am sure I'd forget.

Mind you, this is certainly not very critical, nor a deal breaker. I just wonder about the design choice.

zundertj: That is for me also the biggest design question (see my top comment), and making it recursive is probably the best. I'll work on changing this.

src/Coverage.jl
zundertj: The rule has now become: if the coverage is already empty (`-`), or there is a line exclusion tag, `exclude`. Otherwise, it parses columns 1 to 9.

src/Coverage.jl
zundertj: Done.

src/Coverage.jl
```diff
@@ -201,24 +201,38 @@ module Coverage
...
             end
         end
+
+        # check for excluded lines
+        l = 1
+        let io = IOBuffer(content)
+            while !eof(io)
+                line = readline(io)
+                # Verify that there is no line exclusion tag
+                if any(pat -> occursin(pat, line[10:end]), excluded_line_patterns)
```
vtjnash:
```suggestion
                if any(pat -> occursin(pat, line), excluded_line_patterns)
```

zundertj:
So I have all the test working now locally, but both Travis and AppVeyor keep failing. If anyone has any pointers, that would be much appreciated.

DilumAluthge:
@zundertj Why did you delete the file `test/data/Coverage.jl.cov`? Don’t you need to have that file inside the `test/data/no_exclusions` directory?

DilumAluthge:
@zundertj On an unrelated note, could you edit the original post to have the following content exactly? That way, when this PR is merged, GitHub will automatically close the mentioned issues and pull requests. (But you have to use those specific words like `Fixes` and `Closes` in order for GitHub to understand you.)

> Fixes #44 
> Fixes #53 
> Closes #55

DilumAluthge:
Now that the core functionality has been factored out to CoverageTools.jl, this PR will need to be made to the https://github.com/JuliaCI/CoverageTools.jl repository.

PR #219, opened 2019-03-11
Improve isfuncexpr further using Base.is_short_function_def

fingolfin:


fingolfin:
Turns out `Base.is_short_function_def` is not quite enough either, e.g. it does not handle `() -> x`

test/runtests.jl
```diff
@@ -40,7 +40,9 @@ end
     @test !Coverage.isfuncexpr(:(f = x))
     @test Coverage.isfuncexpr(:(() -> x))
     @test Coverage.isfuncexpr(:(x -> x))
-    @test Coverage.isfuncexpr(:(g() where {T} = 3))
+    @test Coverage.isfuncexpr(:(f() where A = x))
+    @test Coverage.isfuncexpr(:(f() where A where B = x))
+    @test_broken Coverage.isfuncexpr(:(()::C = x))
```
vtjnash: That's because it's trying to defining a variable named (), not a function.

```julia
julia> f()::Int = x
f (generic function with 1 method)

julia> ()::Int = x
ERROR: TypeError: in typeassert, expected Int64, got Tuple{}
Stacktrace:
 [1] top-level scope at REPL[14]:1
```

fingolfin: Ah, OK, thanks for the clarification! I blindly copied this without understanding it (which, as always, of course was a mistake on my part) from your comment on https://github.com/JuliaCI/Coverage.jl/pull/215#discussion_r264065514 (but perhaps I misunderstood that comment)

PR #220, opened 2019-03-11
Remove verbose flag, change some @info to @debug

fingolfin:
There is no point in Coverage.LCOV.readfolder and Coverage.clean_folder
reporting every file they did *not* process; this just clogs up build logs,
and almost always is useless; so change this from @info to @debug.

Also, by default user won't want to know the JSON replies of Codecov and
Coveralls, nor the exact URL we used to submit to Codecov, so change those
from @info to @debug as well.

On the other hand, knowing which files actually are being processed *is*
useful, so add an @info message for that (I recently had an exception thrown
in that code and had a hard time figuring out which file was being proceed, so
this will help).

Finally, get rid of the `verbose` keyword flag; instead, users can adjust the
logging level to suppress or show various messages.

### DIFF
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -182,27 +182,24 @@ module Codecov
     """
     function submit_generic(fcs::Vector{FileCoverage}; kwargs...)
         @assert length(kwargs) > 0
-        dry_run = false
-        verbose = true
-        for (k,v) in kwargs
-            if k == :dry_run
-                dry_run = true
-            end
-            if k == :verbose
-                verbose = v
-            end
+        dry_run = get(kwargs, :dry_run, false)
+        if haskey(kwargs, :verbose)
+            Base.depwarn("The verbose keyword argument is deprecated, set the environment variable " *
+                         "JULIA_DEBUG=Coverage for verbose output", :submit_generic)
+            verbose = kwargs[:verbose]
+        else
+            verbose = false
         end
         uri_str = construct_uri_string(;kwargs...)
 
-        if verbose
-            @info "Codecov.io API URL:\n" * mask_token(uri_str)
-        end
+        verbose && @info "Submitting data to Codecov..."
+        verbose && @debug "Codecov.io API URL:\n" * mask_token(uri_str)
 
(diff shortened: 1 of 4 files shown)


src/codecovio.jl
```diff
@@ -225,7 +215,7 @@ module Codecov
 
         uri_str = "$(codecov_url)/upload/v2?"
         for (k,v) in kwargs
-            if k != :codecov_url && k != :dry_run && k != :verbose
```
ararslan: If the user does pass `verbose` as a keyword, removing the explicit handling here will mean it gets added to the URL string.

src/codecovio.jl
```diff
@@ -182,27 +182,17 @@ module Codecov
     """
     function submit_generic(fcs::Vector{FileCoverage}; kwargs...)
         @assert length(kwargs) > 0
-        dry_run = false
-        verbose = true
```
ararslan: We should implement an informative deprecation, so that people using the `verbose` keyword know that to get the same logging level, they should use the logging framework instead. I'd add something like
```julia
if haskey(kwargs, :verbose)
    Base.depwarn("The verbose keyword argument is deprecated, set the environment variable " *
                 "JULIA_DEBUG=Coverage for verbose output", :submit_generic)
    verbose = kwargs[:verbose]
else
    verbose = false
end
```
then keep all existing logic in place for the duration of the deprecation period. That avoids a hard break for existing code.

fingolfin: Are you aware of any code or people using the "verbose" mechanism?

fingolfin: (Just to clarify, I'll do what you suggest anyway, though, but am still wondering)

src/codecovio.jl
fingolfin: Done, and added a comment

src/codecovio.jl
ararslan: I don't, it's more just a precaution.

fingolfin: @ararslan OK now?

PR #222, opened 2019-03-16
Mask codecov token in logs

zundertj:
Reimplements #113 .

I have taken a different approach to the original PR: instead of splitting the url construction logic I left that as-is, and used a regex on the full string instead. That makes testing a bit easier as well.

One potential problem might be the line
```
            @info "Result of submission:" * String(req)
```
Does this print the url string as well?

### DIFF
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -195,7 +195,7 @@ module Codecov
         uri_str = construct_uri_string(;kwargs...)
 
         if verbose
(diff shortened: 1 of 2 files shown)


fingolfin:
Thanks for working on this! Yes, the `@info` also may print (log) the URL.

fingolfin:
... ah, except I misread and you are talking about the JSON result of the request, and I don't think the token is in there.

PR #223, opened 2019-03-17
add missing bounds check in amend_coverage_from_src

vtjnash:
avoid potential out-of-bounds array access attempt in `function_body_lines`

### DIFF
--- a/src/parser.jl
+++ b/src/parser.jl
@@ -70,7 +70,8 @@ function function_body_lines!(flines, ast::Expr, coverage::Vector{CovCount}, lin
         # speaking, and so we should *not* apply our heuristic (which is mean
         # to mark functions which were never executed as code, which the
         # coverage data returned by Julia normally does not)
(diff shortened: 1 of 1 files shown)


src/parser.jl
```diff
@@ -70,7 +70,8 @@ function function_body_lines!(flines, ast::Expr, coverage::Vector{CovCount}, lin
         # speaking, and so we should *not* apply our heuristic (which is mean
         # to mark functions which were never executed as code, which the
         # coverage data returned by Julia normally does not)
-        if all(l -> coverage[l+lineoffset] === nothing, flines_new)
+        not_covered(l) = (l > length(coverage) || coverage[l + lineoffset] === nothing)
```
fingolfin: shouldn't that be:
```suggestion
        not_covered(l) = (l + lineoffset > length(coverage) || coverage[l + lineoffset] === nothing)
```

PR #239, opened 2019-10-10
Delete REQUIRE

DilumAluthge:
We no longer have any test-only dependencies, so we can delete `test/REQUIRE`.

### DIFF
--- a/test/REQUIRE
+++ b/test/REQUIRE
@@ -1,1 +1,0 @@
-Suppressor 0.1.1


DilumAluthge:
The Travis "failures" on `Travis CI - Branch` and `Travis CI - Pull Request` are due to some problem in the interaction between `staging.travis-ci.com` and Coveralls. It seems that Coveralls does not accept coverage reports when they are submitted from `staging.travis-ci.com`.

The correct Travis builds are `continuous-integration/travis-ci/pr` and `continuous-integration/travis-ci/push` (these correspond to `travis-ci.org` builds). Those Travis builds are passing.

PR #251, opened 2019-11-21
Add pull request number to Coveralls.io

vtjnash:
We send it to CodeCov.io, so just copying it over to Coveralls.


vtjnash:
I can't really tell what Coveralls.io does with this to know that it's correct (I suspect it's been inferring the same information successfully), but should be good to go. Will merge in a bit, unless someone chimes in that they wanted to review first.

PR #270, opened 2020-04-21
Test Travis on Windows/Mac/Linux, test on more Julia versions, drop support for Julia 0.7

omus:
Testing on Travis CI only instead of both Appveyor and Travis. Should make maintaining simpler and should allow the CI to finish faster too.

Also, now we are testing on Julia 0.7 and 1.4 which this package supports. Probably we should drop 0.7 and remove it from the Project.toml.


DilumAluthge:
Yeah drop 0.7.

.travis.yml
```diff
@@ -11,14 +11,31 @@ language: julia
...
+  - x64
+  - x86
 
 julia:
   - "1.0"
   - "1.1"
   - "1.2"
   - "1.3"
+  - "1.4"
+  - "1" # "1" automatically expands to the latest stable release of the form `1.y.z`
```
omus, line 26: Why quote the numbers?

DilumAluthge: There is a bug in YAML where `1.0` turns in to `1`. Which is not the same thing.

So we need to quote `"1.0"`. And for consistency I quote all numbers as a matter of habit. Otherwise it looks quite weird.

DilumAluthge: I think Travis has a beta feature that fixes this issue. But I'd rather not rely on that. Easiest to quote version numbers, that way we ensure it means what we want it to mean.

omus: Are you sure about that? I haven't quoted my `1.0` in TimeZones.jl and https://github.com/JuliaTime/TimeZones.jl/blob/master/.travis.yml#L32 and the CI uses 1.0.1: https://travis-ci.org/github/JuliaTime/TimeZones.jl/jobs/677696077

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Coverage"
 uuid = "a2441757-f6aa-5fb2-8edb-039e3f45d037"
 authors = ["Iain Dunning <iaindunning@gmail.com>"]
-version = "1.0.0"
+version = "1.1.0"
```
omus, line 4: Needs to be a minor bump if we drop Julia 0.7 from compat

.travis.yml
DilumAluthge: It absolutely used to be the case. See e.g. https://github.com/travis-ci/docs-travis-ci-com/pull/1537#issuecomment-342764863

Late last year, Travis started a new beta "Build Config Validation" that allegedly does not have this issue. I'm assuming that's what is happening on TimeZones.jl. I don't really want to have a `.travis.yml` file that relies on a relatively recent feature in Travis. I'd rather just have a `.travis.yml` file that we know will work correctly regardless.

Project.toml
DilumAluthge: Good catch.

DilumAluthge:
@omus Want to squash and rebase?

.travis.yml
omus: Interesting. It's possible I missed when it was broken.

omus: Do we want to include `"1"` as it's currently redundant and will make the CI take longer?

omus:
> Want to squash and rebase?

I was planning on using the GitHub "Squash and merge"

.travis.yml
DilumAluthge: Good question. The advantage of "1" is that if we forget to add "1.5" when 1.5 is released, it will still get tested.

What do you think about taking "1.4" off the list? 1.4 will get tested automatically by the "1" entry.

Then when 1.5 is released, we can add "1.4" to the list.

That way, the consequence of forgetting to update the Travis file is that we don't test on the second most recent version. In my mind, that consequence is not as bad as forgetting to test on the most recent version.

omus: I can get behind that. The only disadvantage I see is that `"1"` sometimes takes a while to get update and can fall behind the latest release. Having it update eventually though is better here.

DilumAluthge:
@omus Let us allow test failures on nightly for now? And then we can disallow failures on nightly once tests are passing on nightly.

.travis.yml
```diff
@@ -11,14 +11,30 @@ language: julia
...
 julia:
   - "1.0"
   - "1.1"
   - "1.2"
   - "1.3"
+  - "1"   # Latest stable release, currently "1.4". Add "1.4" into the matrix when it is no longer the latest.
   - nightly
 
+jobs:
+  fast_finish: true
```
omus, line 29:
```suggestion
  fast_finish: true
  allow_failures:
    - julia: nightly
```

omus:
We'll need to remove Appveyor once this is merged.

omus:
I'll wait until tomorrow to merge so I have availability to sort out the loose ends

omus:
@DilumAluthge I'm not sure if you have admin access to the JuliaCI appveyor group but if you do could you go to: https://ci.appveyor.com/project/JuliaCI/coverage-jl/settings and make sure: "Skip branches without `appveyor.yml`" is checked?

omus:
I think we also need to check:

> Do not build on "Pull request" events

DilumAluthge:
I don't have admin access. @StefanKarpinski

ararslan:
Done. I can just delete the AppVeyor project if it's no longer used.

omus:
We may want to leave it around for a little while in case we want to re-enable it for a reason I'm currently unaware of.

PR #279, opened 2020-08-13
better error msg and support coverall.yml

Roger-luo:
currently on github actions, if the COVERALL_TOKEN is not set as env variable, the error msg will say `ERROR: KeyError: key "service_name" not found` instead of the error msg should be printed on a non-travis CI.

on the other hand, I guess using `.coverall.yml` is perferred by Coverall, given that they provide this instruction on their help page:

image

But currently `Coverage.submit` won't read this file, thus I added support for that.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -9,11 +9,13 @@ HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
 JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
 MbedTLS = "739be429-bea8-5141-9913-cc70e7f3736d"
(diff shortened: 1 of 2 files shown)


vtjnash:
Requests for corrections to the instructions should be posted to https://github.com/lemurheavy/coveralls-public/issues. But that's not the instructions I see for https://docs.coveralls.io/julia:
image


I've updated this code to include an explicit error if service_name is not detected (plus provided an extra environment variable to set it): https://github.com/JuliaCI/Coverage.jl/blob/5f75767166f2e7f51db02da8edc4b2fed7d999f6/src/coveralls.jl#L124-L126

The bolded text that shouts "do not make your repo token public" would often contradict the suggestion to use a .coverall.yml file, so I'm generally inclined not to support that.

Roger-luo:
OK, I'll close this then.

PR #281, opened 2020-09-04
Coveralls Enterprise support

tk3369:
I am trying to integrate with Coveralls Enterprise so I need to hook it up with an on-prem server.  This PR adds supports for two new environment variables:

* `COVERALLS_URL`: Coveralls API endpoint. Defaults to `https://coveralls.io/api/v1/jobs`
* `SSL_VERIFY`: Enable SSL certificate verification during request. Defaults to `1`.

P.S. No unit test has been written as we don't currently have any for the `post_request` call.

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -187,7 +187,10 @@ module Coveralls
     # posts the actual request given the data
     function post_request(data, verbose)
         verbose && @info "Submitting data to Coveralls..."
-        req = HTTP.post("https://coveralls.io/api/v1/jobs", HTTP.Form(makebody(data)))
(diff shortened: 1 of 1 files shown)


fingolfin (approved):

fingolfin:
The CI failures are fixed now. Would you mind rebasing this PR to fix the conflict and so that tests can run again? Thank you!

vtjnash:
I ended up adding a COVERALLS_URL variable already in https://github.com/JuliaCI/Coverage.jl/pull/289/commits/d351511d94019a05d4003866993eeaeebf2d1fdb, as a means of submitting to a mock server (that's where it gets unit tested too). Rather than disable `require_ssl_verification`, I think it might make more sense to use http instead for your local server?

PR #288, opened 2020-11-06
Support Github Actions with Codecov

kyungminlee:
Addresses issue https://github.com/JuliaCI/Coverage.jl/issues/284

PR #289, opened 2020-11-11
updates for v1.5, etc.

### LINKED ISSUE #273: Travis is failing on Julia nightly

vtjnash:
Fixes #273
Fixes https://github.com/JuliaLang/julia/issues/38173

### DIFF
--- a/README.md
+++ b/README.md
@@ -178,17 +178,6 @@ When using Coverage.jl locally, over time a lot of `.cov` files can accumulate. 
        - C:\projects\julia\bin\julia -e "using Pkg; Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"
        ```
 
-4. If you are uploading from multiple jobs, you'll need to tell Coveralls to merge the results after all CI jobs have completed. See https://docs.coveralls.io/parallel-build-webhook for general instructions.
-
-    For Travis, this can be achieved by adding the following to `.travis.yml`:
-
-        env:
-          global:
-            - COVERALLS_PARALLEL=true
-        notifications:
-          webhooks: https://coveralls.io/webhook
-
-
 ## A note for advanced users
 
 Coverage tracking in Julia is not yet quite perfect. One problem is that (at
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,9 @@ notifications:
 script:
   - julia --color=yes --check-bounds=yes etc/travis-test.jl
   - export COVERALLS_PARALLEL=true
-  # submit coverage data, and collect new coverage data on that
-  - julia --color=yes --code-coverage=user etc/travis-coverage.jl
+  # submit coverage data to a black hole server with a fake job-id,
+  # and collect new coverage data on that
+  - COVERALLS_URL=https://httpbingo.julialang.org/post
+    julia --color=yes --code-coverage=user etc/travis-coverage.jl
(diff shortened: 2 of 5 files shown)


DilumAluthge:
FWIW, since Travis is ending its unlimited free tier, it might be worth moving this package's CI to GitHub Actions at some point.

DilumAluthge:
Travis was passing earlier, now it seems to be failing.

vtjnash:
Yeah, I was still testing different settings and did that intentionally. This has many new updates now, and should be good to review.

README.md
```diff
@@ -178,17 +178,6 @@ When using Coverage.jl locally, over time a lot of `.cov` files can accumulate.
...
 
-4. If you are uploading from multiple jobs, you'll need to tell Coveralls to merge the results after all CI jobs have completed. See https://docs.coveralls.io/parallel-build-webhook for general instructions.
-
-    For Travis, this can be achieved by adding the following to `.travis.yml`:
-
-        env:
-          global:
-            - COVERALLS_PARALLEL=true
-        notifications:
-          webhooks: https://coveralls.io/webhook
```
fingolfin, line 189: I am surprised this is not necessary anymore. They still mention it on <https://github.com/marketplace/actions/coveralls-github-action> and <https://docs.coveralls.io/parallel-build-webhook>, and it was absolutely vital when I last played with it -- but that was some time ago; so if that's what you found empirically, let's do it... Coveralls documentation sucks in general, so I am not completely surprised it might not reflect reality anymore :-).

I'll soon need to dig into this again for another project (where Travis cut our legs, so we moved to GitHub Actions, and that broke Coveralls integration), I'll keep this in mind, might simplify things there!

fingolfin (approved):
This PR modifies a lot of things at once. I tried to look at all the changes, and they appear OK, but I won't claim to have checked all of them in detail. They seem broadly sane, though, and if there are regressions, we'll just have to deal with them as they are reported. In the meantime, it seems to unbreak the CI tests and hopefully things in general, so it should be merged.

Thanks!

README.md
vtjnash: Yeah, I noticed that even old builds now are showing up correctly without this flag. But the only effect it was now having here was to prevent any builds from showing up (it was unable to process the webhook now, I'm guessing perhaps due to the deprecation of travis-ci.org)

https://github.com/lemurheavy/coveralls-public/issues/1461#issuecomment-725839058

fingolfin: @vtjnash looking at e.g. https://github.com/JuliaCI/Coverage.jl/pull/288 there are now four coveralls comments there. I vaguely recall from way back that this was caused by the "parallel" support not being activated -- but it's been some time, so I might totally misremember!

vtjnash: It used to be disabled entirely, but I'm not sure why it'd be 4 and not 8 (the number of builds) or 1 (the top comment has 5 rolling edits as the parallel builds finished)

fingolfin: Top comment was created, then edited 5 times; plus 3 more comments; means a total of 1+5+3=9 results reported. The Travis build has 9 jobs. So what I think happens is that whenever one of the jobs ends, Coveralls adds or updates an issue comment. Why it didn't just put all updates in one comment, I can only guess at; most likely a race condition, but w/o being able to see the precise timestamps when each comment was made/updated, it's hard to say...

PR #296, opened 2020-11-17
Expand on data extraction from github action runs for Coveralls

mfherbst:
Based on the [official docs](https://docs.github.com/en/free-pro-team@latest/actions/reference/environment-variables).

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -102,8 +102,14 @@ module Coveralls
                 data["git"]["branch"] = split(ENV["GIT_BRANCH"], "/")[2]
             end
         elseif haskey(ENV, "GITHUB_ACTION")
+            data["service_job_id"] = ENV["GITHUB_RUN_ID"]
             data["service_name"] = "github"
             data["git"] = parse_git_info(git_info)
+
+            event_path = open(JSON.Parser.parse, ENV["GITHUB_EVENT_PATH"])
+            github_pr_info = get(event_path, "pull_request", Dict())
+            github_pr = get(github_pr_info, "number", "")
+            isempty(github_pr) || (data["service_pull_request"] = github_pr)
         else
             data["git"] = parse_git_info(git_info)
         end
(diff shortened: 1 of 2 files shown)


mfherbst:
Regarding tests ... it's not quite clear to me how to cover this. Happy for ideas.

src/coveralls.jl
```diff
@@ -102,8 +102,14 @@ module Coveralls
...
             end
         elseif haskey(ENV, "GITHUB_ACTION")
+            data["service_job_id"] = ENV["GITHUB_RUN_ID"]
             data["service_name"] = "github"
             data["git"] = parse_git_info(git_info)
+
+            event_path = open(JSON.Parser.parse, ENV["GITHUB_EVENT_PATH"])
+            github_pr_info = get(event_path, "pull_request", Dict())
+            github_pr = get(github_pr_info, "number", "")
+            isempty(github_pr) || (data["service_pull_request"] = github_pr)
```
fingolfin, line 112: Ooooh, that's nice, I wasn't aware of it, else I would have suggested it for PR #288 -- I wonder if perhaps you or @kyungminlee would be interested in adjusting the Codecov code path to also parse the `GITHUB_EVENT_PATH` JSON. In general, we try to keep the Codecov and Coveralls support in sync (but I merged PR #288 anyway, because (a) I forgot 😨 and (b) honestly, I'd rather have something that works in now than to not have it at all because it's perfect...

Anyway: very nice 😄

fingolfin (approved):
Thanks!

It'd be nice to have some tests for this, just like we have for the other combinations of CI provider and Codecov/Coveralls in `test/runtests.jl`.

But I personally wouldn't want to hold this PR hostage over this :-)

src/coveralls.jl
mfherbst: Sure I can do that.

mfherbst: It's a bit tricky to use in general though, because for push builds you don't have that much info (like I could not find how to e.g. extract the branch name in a quick search).

fingolfin: I guess one then needs to combine techniques from PR #288 and this one... Ah well.

mfherbst:
:+1: Thanks for merging.

PR #304, opened 2021-02-13
Update Codecov upload API to v4

ararslan:
Guessing based on https://docs.codecov.io/reference#upload.

### DIFF
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -227,7 +227,11 @@ module Codecov
             # Tell Codecov we have an upload for them
             response = HTTP.post(uri_str; headers=Dict("Accept" => "text/plain"))
             # Get the temporary URL to use for uploading to S3
-            s3url = split(String(response.body), '\n')[2]
+            repr = String(response)
+            s3url = get(split(String(response.body), '\n'), 2, "")
+            repr = chomp(replace(repr, s3url => ""))
+            @debug "Result of submission:" * repr
+            startswith(s3url, "https://") || error("Invalid codecov response: $s3url")
             # Upload to S3
             request = HTTP.put(s3url; body=json(to_json(fcs)),
                                header=Dict("Content-Type" => "application/json",
--- a/src/codecovio.jl
+++ b/src/codecovio.jl
@@ -224,10 +224,15 @@ module Codecov
         @debug "Codecov.io API URL:\n" * mask_token(uri_str)
 
         if !dry_run
-            heads   = Dict("Content-Type" => "application/json")
-            data    = to_json(fcs)
-            req     = HTTP.post(uri_str; body = JSON.json(data), headers = heads)
-            @debug "Result of submission:" * mask_token(String(req))
+            # Tell Codecov we have an upload for them
+            response = HTTP.post(uri_str; headers=Dict("Accept" => "text/plain"))
+            # Get the temporary URL to use for uploading to S3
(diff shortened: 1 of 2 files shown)


ararslan:
RIP, looks like there's no CI here anymore. Guess we'll need to set up GHA.

src/codecovio.jl
```diff
@@ -222,11 +222,15 @@ module Codecov
...
         if !dry_run
-            heads   = Dict("Content-Type" => "application/json",
-                           "Accept" => "application/json")
-            data    = to_json(fcs)
-            req     = HTTP.post(uri_str; body = JSON.json(data), headers = heads)
-            @debug "Result of submission:" * String(req)
+            # Tell Codecov we have an upload for them
+            response = HTTP.post(uri_str; headers=Dict("Accept" => "text/plain"))
+            # Get the temporary URL to use for uploading to S3
+            s3url = split(String(response.body), '\n')[2]
```
vtjnash, line 228:
```suggestion
            s3url = get(split(String(response.body), '\n'), 2, "")
            @debug "Result of submission:" * chomp(replace(String(req), s3url => ""))
            startswith(s3url, "https://") || error("Invalid codecov response: $s3url")
```

ararslan: `req` doesn't exist here, I think you meant `response`? Also, won't `String(response.body)` take ownership of the body and leave `String(response)` not informative, or does the `String` method show headers and whatnot?

vtjnash: ah, true, it does add headers, but the other call does destroy the body
```suggestion
            repr = String(response)
            s3url = get(split(String(response.body), '\n'), 2, "")
            repr = chomp(replace(repr, s3url => ""))
            @debug "Result of submission:" * repr
            startswith(s3url, "https://") || error("Invalid codecov response: $s3url")
```

vtjnash:
Can you rebase and re-enable CI submission on https://github.com/JuliaCI/Coverage.jl/blob/master/etc/travis-coverage.jl#L4

vtjnash:
You'll need to remove the conditional on that line. Looks like you might need to disable the blackhole server that we use for testing also (or provide a way to specify the API via an environment variable, and use v2 there?)

DilumAluthge:
I have rebased, and I have enabled submission (i.e. we've removed all conditionals from https://github.com/JuliaCI/Coverage.jl/blob/master/etc/travis-coverage.jl), but it looks like CI is still failing.

vtjnash:
Yeah, it fails on the blackhole server we used for testing.

DilumAluthge:
@ararslan @vtjnash Based on #316, it seems like the only issue with this PR is the blackhole server bug. How do y'all want to proceed?

ararslan:
¯\\\_(ツ)\_/¯

DilumAluthge:
@vtjnash Can you explain the purpose of the black-hole server?

vtjnash:
It avoided cluttering the list of log files with duplicate entries–no practical benefit to it

DilumAluthge:
Alright, I think I have a way to fix it. Basically:
1. If we are not using the "black hole" server, we upload to S3
2. If we are using the "black hole" server, we do not upload to S3

See PR #318.

For the diff between this PR and PR #318, see here: https://github.com/JuliaCI/Coverage.jl/compare/aa/codecov-v4...dpa/codecov-v4

PR #313, opened 2021-06-09
Trying to debug Coveralls

DilumAluthge:


src/coveralls.jl
```diff
@@ -42,9 +42,17 @@ module Coveralls
...
         name = Sys.iswindows() ? replace(fc.filename, '\\' => '/') : fc.filename
-        return Dict("name"          => name,
-                    "source_digest" => digest(MD_MD5, fc.source, "secret"),
-                    "coverage"      => fc.coverage)
+        T_K = String
+        T_V1 = String
+        T_V2 = Dict
+        T_V3 = Vector{<:Dict}
+        T_V4 = Vector{<:Union{Nothing, Int}}
+        T_V = Union{T_V1, T_V2, T_V3, T_V4}
```
vtjnash, line 50: I would heavily discourage you from making Union types explicitly. If you think you want one, you probably just want to use `Any`

DilumAluthge: So, for debugging, I did it to figure out where the non-string values were coming from. E.g. with this `Union` type, I was able to figure out the lines where we tried to assign an `Int` or a `Vector{UInt8}` as a value in the dict.

Once I've gotten everything working, I think it makes sense to just turn everything back to `Dict`, right?

vtjnash: Being explicit about `Dict{String,Any}` might be be bad either, since that is the type JSON expects (and will eventually  need to convert it to, if you initially construct anything else)

DilumAluthge: Yeah I think that's a good call.

PR #310, opened 2021-05-10
Re-enable Coveralls submission from CI

vtjnash:

PR #315, opened 2021-06-09
Fix Coveralls submission, and re-enable Coveralls submission from CI

### LINKED ISSUE #314: Debugging the Coveralls errors

DilumAluthge:
Fixes #314 
Closes #310 
Closes #313

### DIFF
--- a/src/coveralls.jl
+++ b/src/coveralls.jl
@@ -43,7 +43,7 @@ module Coveralls
     function to_json(fc::FileCoverage)
         name = Sys.iswindows() ? replace(fc.filename, '\\' => '/') : fc.filename
         return Dict("name"          => name,
-                    "source_digest" => digest(MD_MD5, fc.source, "secret"),
+                    "source_digest" => bytes2hex(digest(MD_MD5, fc.source, "secret")),
                     "coverage"      => fc.coverage)
     end
 
@@ -102,7 +102,7 @@ module Coveralls
             event_path = open(JSON.Parser.parse, ENV["GITHUB_EVENT_PATH"])
             github_pr_info = get(event_path, "pull_request", Dict())
             github_pr = get(github_pr_info, "number", "")
-            isempty(github_pr) || (data["service_pull_request"] = github_pr)
+            isempty(github_pr) || (data["service_pull_request"] = strip(string(github_pr)))
         else
             data["git"] = parse_git_info(git_info)
         end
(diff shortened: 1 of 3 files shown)


src/coveralls.jl
```diff
@@ -102,7 +110,9 @@ module Coveralls
             event_path = open(JSON.Parser.parse, ENV["GITHUB_EVENT_PATH"])
             github_pr_info = get(event_path, "pull_request", Dict())
             github_pr = get(github_pr_info, "number", "")
-            isempty(github_pr) || (data["service_pull_request"] = github_pr)
+            github_pr_string = String(strip(_stringify_string_or_int(github_pr)))
+            github_pr_string::String
+            isempty(github_pr) || (data["service_pull_request"] = github_pr_string)
```
vtjnash:
```suggestion
            isempty(github_pr) || (data["service_pull_request"] = strip(string(github_pr)))
```

src/coveralls.jl
```diff
@@ -92,7 +100,7 @@ module Coveralls
 
             # get the name of the branch if not a pull request
             if get(ENV, "CI_PULL_REQUEST", "false") == "false"
-                data["git"]["branch"] = split(ENV["GIT_BRANCH"], "/")[2]
+                data["git"]["branch"] = String(split(ENV["GIT_BRANCH"], "/")[2])
```
vtjnash, line 103: I expect the JSON formatting should already handle AbstractString, SubString, and String the same

src/coveralls.jl
DilumAluthge: Oops I just pushed something that made this review "outdated", so I can't click the "commit suggestion" button, but I'll make that change manually and commit it.

vtjnash (approved):
Thanks for tracking that down!

src/coveralls.jl
```diff
@@ -66,6 +66,14 @@ module Coveralls
...
     end
 
+    function _stringify_string_or_int(str::AbstractString)
+        return str
+    end
+
+    function _stringify_string_or_int(int::Integer)
+        return string(int)
+    end
+
```
vtjnash, line 76:
```suggestion
```

src/coveralls.jl
DilumAluthge: Earlier I was running into issues when it was a `SubString`, but I was debugging, so I had changed a bunch of stuff. Let me get rid of the `String()` here and let's see if it still works.

DilumAluthge:
> Thanks for tracking that down!

Your tip to look at https://github.com/lemurheavy/coveralls-public/issues/1527 was spot-on. The issue was indeed that our JSON was not typed the way that Coveralls expected it to be typed.

src/coveralls.jl
DilumAluthge: Done.

PR #318, opened 2021-06-13
Update Codecov upload API to v4 (replacement for #304)

DilumAluthge:
Closes #304 

Here is this diff for this PR versus PR #304: https://github.com/JuliaCI/Coverage.jl/compare/aa/codecov-v4...dpa/codecov-v4

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -40,10 +40,13 @@ jobs:
           arch: ${{ matrix.arch }}
       - uses: julia-actions/julia-buildpkg@v1
       - uses: julia-actions/julia-runtest@v1
+        env:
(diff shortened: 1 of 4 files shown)


DilumAluthge:
@vtjnash @ararslan Take a look at this PR and let me know what you think.

DilumAluthge:
Also, changing the version of the Codecov API that we use - is that a breaking change or a nonbreaking change?

ararslan:
It's not really exposed to the user in terms of API, so IMO not breaking

DilumAluthge:
> It's not really exposed to the user in terms of API, so IMO not breaking

Sounds good to me!

DilumAluthge:
Alright, let's give this a try.

PR #327, opened 2021-10-15
Rephrase instructions for local usage

timholy:
The rewords some of the instructions in the README for local coverage measurement. The major changes & their motivations are:

- indicate the purpose of each step before describing the mechanics of the step. This also makes it easier to indicate that one does not necessarily need to measure summary statistics if one's only goal is to find lines lacking coverage.
- prioritize `Pkg.test(; coverage=true)` above manual measurement. Most users will want the simpler option.
- clarify the role of the `.info` tracefile. This allows users to determine when they need to supply the more complex set of options. By presenting it as a choice it also makes it clearer that you don't need to execute both lines.
- acknowledge limitations in the accuracy of Julia's coverage analysis

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,22 +22,25 @@ Most users will want to use [Coverage.jl](https://github.com/JuliaCI/Coverage.jl
 
 ### Code coverage
 
-*Step 1:* Navigate to your test directory, and run julia with the `--code-coverage` option:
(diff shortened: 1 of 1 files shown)


timholy:
This is for a class I'm teaching, and the homework gets assigned today, so I'll merge.

PR #192, opened 2019-01-04
WIP: Allow coverage submission from Cirrus CI

ararslan:
This will permit coverage submission from builds running on [Cirrus CI](https://cirrus-ci.org), which will fix the coverage step in [my Cirrus template](https://github.com/ararslan/CirrusCI.jl).

Currently a work in progress as I have no idea if this works.

PR #334, opened 2022-08-13
Allow Codecov submission from Cirrus CI

ararslan:
Picking up from #192, as Cirrus seems now to be better supported by Codecov itself (e.g. https://community.codecov.com/t/add-support-of-uploading-from-cirrus-ci-without-token/1028/34). No idea whether it'll work now but worth another try.

### DIFF
--- a/.cirrus.yml
+++ b/.cirrus.yml
@@ -1,0 +1,27 @@
+freebsd_instance:
+  image: freebsd-12-2-release-amd64
+task:
+  name: FreeBSD
+  env:
+    matrix:
+      - JULIA_VERSION: 1.6
+      - JULIA_VERSION: 1
+      - JULIA_VERSION: nightly
+  allow_failures: $JULIA_VERSION == 'nightly'
+  install_script:
+    - sh -c "$(fetch https://raw.githubusercontent.com/ararslan/CirrusCI.jl/master/bin/install.sh -o -)"
+  build_script:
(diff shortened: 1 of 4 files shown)


ararslan:
Julia 1.0 failure is because I used `Sys.isfreebsd`, which was added in 1.1. Cirrus failures are a bad request:
```
[ Info: Submitting data to Codecov...
LoadError: HTTP.Exceptions.(, "POST, ula/4bac=acirsrdxpl_eus=3&evc=irsbid69929219&omt=ce442eef225b959872dsu=uiC/oerg.l"HTTP.Messages:"""HTTP/1.1 400 Bad Request
server: envoy
content-type: text/html; charset=utf-8
access-control-allow-headers: Origin, Content-Type, Accept, X-User-Agent
allow: GET, POST, HEAD, OPTIONS
content-length: 26
referrer-policy: same-origin
Via: 1.1 google

""")
```

ararslan:
Is this package actually still used now that most packages are using GitHub Actions for CI? Seems Codecov submission is typically handled on GHA by the official Codecov action. I wonder if I can just use the [Codecov uploader](https://docs.codecov.com/docs/codecov-uploader) with the Linux binary plus Linux emulation on Cirrus...

ararslan:
The answer is yes. https://github.com/ararslan/CirrusCI.jl/pull/11

PR #339, opened 2023-04-29
updated README to update the after test path to julia bin

pranavvatsa:
The previous path for the bin folder was giving an error when building. So the documentation now contains the updated path which is compactible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -198,7 +198,7 @@ When using Coverage.jl locally, over time a lot of `.cov` files can accumulate. 
 
        ```yml
        after_test:
-       - C:\projects\julia\bin\julia -e "using Pkg; Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -198,7 +198,7 @@ When using Coverage.jl locally, over time a lot of `.cov` files can accumulate.
 
        ```yml
        after_test:
-       - C:\projects\julia\bin\julia -e "using Pkg; Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"
+       - C:\julia\bin\julia -e "using Pkg; Pkg.add(\"Coverage\"); using Coverage; Coveralls.submit(process_folder())"
```
fingolfin, line 201: I don't know anything about Julia. What is the current (or past) default install directory ?

fingolfin: Oh wait, now I realize this is about AppVeyorand in fact a duplicate of PR #329

Issue #340, opened 2023-08-08
500 Error on Coveralls.submit for Julia 1.9 and nightly


bluesmoon:
Hi,

I have my project (https://github.com/bluesmoon/CurlHTTP.jl) run through travis with Julia 1.6, 1.9 and nightly.  I have the coveralls integration set up (I have tried both coveralls: true as well as the after_success script).

On Julia 1.6 th Coveralls.submit works with no issue
On Julia 1.9 and Julia nightly it fails with a HTTP 500 error.

What should I look at?

PR #349, opened 2025-08-13
Modernize

IanButterworth:
Modernizes Coverage.jl to work with official uploaders from Codecov and Coveralls, replacing the deprecated direct upload functionality. The modernization maintains compatibility.

Developed with Claude.


.github/workflows/CompatHelper.yml
```diff
@@ -1,16 +1,50 @@
...
+        with:
+          version: '1'
+          arch: ${{ runner.arch }}
+        if: steps.julia_in_path.outcome != 'success'
+      - name: "Add the General registry via Git"
+        run: |
+          import Pkg
+          ENV["JULIA_PKG_SERVER"] = ""
+          Pkg.Registry.add("General")
+        shell: julia --color=yes {0}
```
vtjnash, line 28: This sounds slow?
```suggestion
```

IanButterworth: Yeah. This is copied from the CompatHelper example.

.github/workflows/CompatHelper.yml
```diff
@@ -1,16 +1,50 @@
...
+      - name: "Run CompatHelper"
+        run: |
+          import CompatHelper
+          CompatHelper.main()
+        shell: julia --color=yes {0}
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: julia -e 'using CompatHelper; CompatHelper.main()'
+          # This repo uses Documenter, so we can reuse our [Documenter SSH key](https://documenter.juliadocs.org/stable/man/hosting/walkthrough/).
```
vtjnash: Not true. I can add a key though if needed for COMPATHELPER_PRIV as it describes at https://juliaregistries.github.io/CompatHelper.jl/dev/#Creating-SSH-Key

IanButterworth: I just blindly copied this file over. Feel free to push and correct it.

MIGRATION.md
```diff
@@ -0,0 +1,156 @@
...
+### Deprecation Warnings
+If you see deprecation warnings, update your code:
+```julia
+# Old
+Codecov.submit(fcs)
+
+# New
+using Coverage.CIIntegration
+upload_to_codecov(fcs)
+```
```
vtjnash: Seems like we could just tell Claude to rename `CIIntegration.upload_to_codecov` as `Codecov.submit`

README.md
```diff
@@ -1,22 +1,67 @@
...
+The package now provides:
+- 🔄 **Coverage data processing** using CoverageTools.jl
+- 📤 **Export functionality** for official uploaders
+- 🚀 **Automated upload helpers** for CI environments
+- 📋 **Helper scripts** for easy integration
+
+> [!IMPORTANT]
+> **Codecov and Coveralls have deprecated 3rd party uploaders.** Coverage.jl now integrates with their official uploaders while maintaining the same easy-to-use interface for Julia projects.
+>
+> **Migration required:** See [MIGRATION.md](MIGRATION.md) for upgrading from the old `Codecov.submit()` and `Coveralls.submit()` functions.
```
vtjnash: See if Claude can keep the old functions, and downgrade `important` here to just a note that this uses official uploaders

scripts/upload_codecov.jl
```diff
@@ -0,0 +1,168 @@
...
+Examples:
+    julia scripts/upload_codecov.jl
+    julia scripts/upload_codecov.jl --folder src --format lcov --flags julia
+    julia scripts/upload_codecov.jl --dry-run
+"""
+
+using Coverage
+using Coverage.CIIntegration
+
+function parse_args(args)
```
vtjnash: Maybe worth asking Claude to use ArgParse or ArgMacros or whatever is commonly recommended these days. This looks like a functional but very primitive getopt implementation.

scripts/upload_codecov.jl
```diff
@@ -0,0 +1,168 @@
...
+        if success
+            println("🎉 Successfully uploaded to Codecov!")
+            exit(0)
+        else
+            println("❌ Failed to upload to Codecov")
+            exit(1)
+        end
+
+    catch e
+        println("❌ Error: $(string(e))")
```
vtjnash: This will usually display fairly poorly. Either remove try/catch or at least use `sprint(Base.display_error, e)` instead of `string(e)`

src/codecov_export.jl
```diff
@@ -0,0 +1,280 @@
+# Export functionality for Codecov official uploader
+export CodecovExport
+
+"""
+Coverage.CodecovExport Module
+
+This module provides functionality to export coverage data in formats compatible with
+the official Codecov uploader. It replaces the deprecated direct upload functionality.
+"""
+module CodecovExport
```
vtjnash: It looks like you can ask Claude to delete `module Codecov` and replace it with the new functionality `module CodecovExport` renamed to `Codecov`

src/codecov_export.jl
```diff
@@ -0,0 +1,280 @@
...
+
+    @info "Downloading Codecov uploader for $platform..."
+
+    try
+        # Download the uploader
+        Downloads.download(uploader_info.url, exec_path)
+
+        # Make executable on Unix systems
+        if platform != :windows
+            chmod(exec_path, 0o755)
```
vtjnash:
```suggestion
            chmod(exec_path, 0o555)
```
R-X seems sufficient (I think you should be able to set this on all platforms, and some Windows configurations will require it)

src/codecov_export.jl
```diff
@@ -0,0 +1,280 @@
...
+    if auto_download
+        @info "Codecov uploader not found, downloading..."
+        return download_codecov_uploader(; install_dir=install_dir)
+    else
+        error("Codecov uploader not found. Set auto_download=true or install manually.")
+    end
+end
+
+"""
+    generate_codecov_yml(; flags=nothing, name=nothing, output_file="codecov.yml")
```
vtjnash: What's the point of this? Seems complicated to have our entire own yaml system here that is unused and mostly untested

src/coverage_utils.jl
```diff
@@ -0,0 +1,170 @@
...
+"""
+    download_with_info(url::String, dest_path::String, binary_name::String, platform::Symbol)
+
+Download a binary with standardized info messages.
+"""
+function download_with_info(url::String, dest_path::String, binary_name::String, platform::Symbol)
+    @info "Downloading $(binary_name) for $(platform)..."
+    Downloads.download(url, dest_path)
+    chmod(dest_path, 0o755)  # Make executable
+    @info "$(binary_name) downloaded to: $(dest_path)"
```
vtjnash: Might be worth asking Claude to try to DRY its PR, since I think I saw this helper written elsewhere too

src/coverage_utils.jl
```diff
@@ -0,0 +1,170 @@
+# Common utilities for Coverage.jl modules
+module CoverageUtils
```
vtjnash, line 2: Not certain what Claude thought was common in this file, since it seems to not use much from it :joy:

src/ci_integration.jl
```diff
@@ -0,0 +1,275 @@
+"""
+CI Integration helpers for Coverage.jl
+
+This module provides convenience functions for integrating Coverage.jl with
+Continuous Integration platforms using official uploaders.
+"""
+module CIIntegration
```
vtjnash: I don't think we really need each file to be a separate module

src/coveralls_export.jl
```diff
@@ -0,0 +1,368 @@
...
+using SHA
+using LibGit2
+using ..CoverageUtils
+
+export prepare_for_coveralls, export_coveralls_json, download_coveralls_reporter, get_coveralls_executable
+
+# Platform-specific Coveralls reporter installation methods
+const COVERALLS_REPORTERS = Dict(
+    :linux => (
+        url = "https://github.com/coverallsapp/coverage-reporter/releases/latest/download/coveralls-linux",
```
vtjnash:
```suggestion
        url = "https://github.com/coverallsapp/coverage-reporter/releases/latest/download/coveralls-linux-$arch",
```
where `const arch = Sys.ARCH == :aarch64 ? "aarch64" : "x86_64"`

src/codecov_export.jl
```diff
@@ -0,0 +1,280 @@
...
+using ..CoverageUtils
+
+export prepare_for_codecov, export_codecov_json, download_codecov_uploader, get_codecov_executable
+
+# Platform-specific codecov uploader URLs and checksums
+const CODECOV_UPLOADERS = Dict(
+    :linux => (
+        url = "https://uploader.codecov.io/latest/linux/codecov",
+        checksum = nothing  # Will be fetched dynamically
+    ),
```
vtjnash:
```suggestion
    :linux => (
        url = "https://uploader.codecov.io/latest/linux/codecov",
        checksum = nothing  # Will be fetched dynamically
    ),
    :aarch64 => (
        url = "https://uploader.codecov.io/latest/aarch64/codecov",
        checksum = nothing
    ),
```

src/coverage_utils.jl
IanButterworth: It's more common now.

IanButterworth:
I haven't checked that this is actually backwards compatible, but the main functionality seems to be there.

We could cut a 2.0 and just rm all the old stuff? @vtjnash what do you think?

IanButterworth:
To be more specific, I haven't double-checked that existing CI on other repos still work fully with this.

vtjnash:
I think people noted this just literally has done nothing for over 2 years (https://github.com/JuliaCI/Coverage.jl/issues/340), so I think we can just change it to something that works and make a new minor release so that people get something that mostly works, even if it works a bit different

IanButterworth:
Trying this out here https://github.com/JuliaCI/julia-buildkite/pull/475

IanButterworth:
A wrinkle here is that at least our macOS CI runners don't let you run homebrew, and Coveralls don't provide macOS binaries directly. Not sure what we can do for macOS.

vtjnash:
The CI here is also failing (`#define EBADARCH        86      /* Bad CPU type in executable */`)

IanButterworth:
Thanks @vtjnash 👍

IanButterworth:
Did I need to change something here? Seems to have not worked? https://buildkite.com/julialang/julia-buildkite-scheduled/builds/1397#0198afd5-474a-42a2-a2f8-9bdc4ded947d

vtjnash:
It is just using the wrong brew, since Claude first checks the path. Honestly we possibly should just add this to Yggy, since their build system and brew is quite a mess and compiles down to a half a gigabyte installer. We could likely do far better in Julia than their crystal/zig/docker/brew build stack for their ruby script

PR #362, opened 2026-09-12
Support HTTP.jl 2

asinghvi17:
Moves the HTTP.jl compat bound to `2` and migrates the call sites that 2.0 changed. `julia` compat moves to `1.10`, the floor HTTP.jl 2.0 requires.

Verified locally on Julia 1.13: **372 passing with HTTP 2.6.7**, identical to the HTTP 1.11 baseline on `master`.

## Changes

| Site | Change | Reason |
|---|---|---|
| `Codecov.submit_generic`, `Codecov.upload_to_s3`, `Coveralls.post_request` | `String(msg)` → `string(msg)` | 2.0 drops `Base.String(::Message)`. `string` yields the same status-line/headers/body dump these `@debug` statements already logged. |
| `Codecov.upload_to_s3` | dropped the `header=` keyword | 2.0 raises `MethodError` on unrecognized request keywords (details below). |
| `test/runtests.jl` | `JSON.parse(request.body)` → `JSON.parse(String(request.body))` | Server-side request bodies are now `HTTP.BytesBody` rather than a byte vector. |

Untouched by the migration: `HTTP.post`/`HTTP.put` verb calls, `HTTP.Form` and `HTTP.Multipart` uploads, the `query` and `status_exception` keywords, `HTTP.serve!`/`HTTP.port`/`HTTP.header`, and `String(response.body)`. Each behaves the same under 2.0.

## On the dropped `header=` keyword

`HTTP.put` has never recognized `header` — the keyword is `headers` — so 1.x silently discarded the dict, and the S3 upload has always gone out without `Content-Type` or `x-amz-storage-class`. HTTP.jl 2.0 rejects unknown keywords, which is what surfaced this.

This PR keeps the existing wire behavior instead of renaming the keyword to `headers`, because the target is an S3 presigned PUT: an `x-amz-storage-class` header outside the presigned `X-Amz-SignedHeaders` set makes S3 reject the upload with `SignatureDoesNotMatch`. Sending those headers is a separate change that also needs the signing side to cover them — happy to split it out if you'd like them enabled.

This path sits in `submit_generic`, which is already deprecated in favor of the Codecov uploader binary.

## Notes for review

- `julia = "1.10"` is required rather than cosmetic: HTTP.jl 2.0 declares `julia = "1.10.0 - 1"
- This drops HTTP 0.8/0.9/1 support
- [HTTP.jl 1.x → 2.0 migration guide](https://github.com/JuliaWeb/HTTP.jl/blob/master/docs/src/guides/migration-1x.md)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


DilumAluthge (approved):

DilumAluthge:
CI is sad?

asinghvi17:
The CI issues are all coming from Reseau.jl it seems, cc @quinnj

